PR #3, opened 2024-12-16
feat(registry): Create v0 Registry contract

finnian0826:
Registry contract created with:

Write methods:
- `registerReferrer` (owner only)
- `registerReferral`

Read methods:
- `getReferrers` for a given protocol
- `getUsers` for a given protocol, referrer pair
- `getRewardRate` for a given protocol, referrer pair

`npx hardhat compile` to create the ABI (I cleaned up output to what it is now)

Unit tests added

### DIFF
--- a/contracts/Registry.sol
+++ b/contracts/Registry.sol
@@ -81,9 +81,9 @@ contract Registry is AccessControlDefaultAdminRules {
     string calldata protocolId
   ) external {
     // Check if the referrer has been initialized but the user has not been registered before to the given protocol
-    if (_referrers[protocolId][referrerId].rewardAddress != address(0)) {
+    if (_referrers[protocolId][referrerId].rewardAddress == address(0)) {
       revert ReferrerNotRegistered(protocolId, referrerId);
-    } else if (_usersByProtocol[protocolId][msg.sender].timestamp == 0) {
+    } else if (_usersByProtocol[protocolId][msg.sender].timestamp != 0) {
       revert UserAlreadyRegistered(protocolId, referrerId, msg.sender);
     } else {
       _usersByReferrer[protocolId][referrerId][msg.sender] = User(
--- a/contracts/Registry.sol
+++ b/contracts/Registry.sol
@@ -59,7 +59,7 @@ contract Registry is AccessControlDefaultAdminRules {
     uint256[] calldata rewardRates,
     address rewardAddress
   ) external onlyRole(DEFAULT_ADMIN_ROLE) {
-    // Remove referrer from protocols where the relationship is no longer active (from previous registration)
+    // Remove referrer from protocols from previous registrations
     for (uint256 i = 0; i < _referrerIdToProtocolIds[referrerId].length; i++) {
       string memory protocol = _referrerIdToProtocolIds[referrerId][i];
       string[] storage referrerIds = _protocolIdToReferrerIds[protocol];
--- a/contracts/Registry.sol
+++ b/contracts/Registry.sol
@@ -1,0 +1,114 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import {AccessControlDefaultAdminRules} from '@openzeppelin/contracts/access/extensions/AccessControlDefaultAdminRules.sol';
+import {SafeERC20} from '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
+import {IERC20} from '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+
+contract Registry is AccessControlDefaultAdminRules {
+  using SafeERC20 for IERC20;
+
+  event ReferrerRegistered(
+    string indexed referrerId,
+    string[] indexed protocolIds,
+    uint256[] rewardRates,
+    address rewardAddress
+  );
+
+  event ReferralRegistered(
+    string indexed protocolId,
+    string indexed referrerId,
+    address indexed userAddress
+  );
+  
+  event ReferralSkipped(
+    string indexed protocolId,
+    string indexed referrerId,
+    address indexed userAddress
+  );
+
+  struct User {
+    uint256 timestamp;
+  }
+
+  struct Referrer {
+    address rewardAddress;
+    uint256 rewardRate;
+    address[] userAddresses;
+  }
+
+  struct Protocol {
+    string[] _referrers;
+  }
+
+  mapping(string => Protocol) private _protocols;
+  mapping(string => mapping(string => Referrer)) private _referrers;
+  mapping(string => mapping(string => mapping(address => User))) private _users;
+
+  constructor(
+    address owner,
+    uint48 transferDelay
+  ) AccessControlDefaultAdminRules(transferDelay, owner) {}
+
+  function registerReferrer(
+    string calldata _referrerId,
+    string[] calldata _protocolIds,
+    uint256[] calldata _rewardRates,
+    address _rewardAddress
+  ) external onlyRole(DEFAULT_ADMIN_ROLE) {
+    for (uint256 i = 0; i < _protocolIds.length; i++) {
+      _referrers[_protocolIds[i]][_referrerId] = Referrer(
+        _rewardAddress,
+        _rewardRates[i],
+        new address[](0)
+      );
+      _protocols[_protocolIds[i]]._referrers.push(_referrerId);
+    }
+    emit ReferrerRegistered(_referrerId, _protocolIds, _rewardRates, _rewardAddress);
+  }
+
+  function registerReferral(
+    string calldata referrerId,
+    string calldata protocolId
+  ) external {
+    if (
+      // Check if the referrer has been initialized but the user has not been registered before
+      _referrers[protocolId][referrerId].rewardAddress != address(0) &&
+      _users[protocolId][referrerId][msg.sender].timestamp == 0
+    ) {
+      _users[protocolId][referrerId][msg.sender] = User(block.timestamp);
+      _referrers[protocolId][referrerId].userAddresses.push(msg.sender);
+      emit ReferralRegistered(protocolId, referrerId, msg.sender);
+    } else {
+      emit ReferralSkipped(protocolId, referrerId, msg.sender);
+    }
+  }
+
+  function getReferrers(
+    string calldata protocolId
+  ) external view returns (string[] memory) {
+    return _protocols[protocolId]._referrers;
+  }
+
+  function getUsers(
+    string calldata protocolId,
+    string calldata referrerId
+  ) external view returns (address[] memory, uint256[] memory) {
+    address[] memory userAddresses = _referrers[protocolId][referrerId]
+      .userAddresses;
+    uint256[] memory timestamps = new uint256[](userAddresses.length);
+
+    for (uint256 i = 0; i < userAddresses.length; i++) {
+      timestamps[i] = _users[protocolId][referrerId][userAddresses[i]].timestamp;
+    }
+
+    return (userAddresses, timestamps);
+  }
+
+  function getRewardRate(
+    string calldata protocolId,
+    string calldata referrerId
+  ) external view returns (uint256) {
+    return _referrers[protocolId][referrerId].rewardRate;
+  }
+}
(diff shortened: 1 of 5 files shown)


contracts/Registry.sol
```diff
@@ -0,0 +1,114 @@
...
+        _rewardAddress,
+        _rewardRates[i],
+        new address[](0)
+      );
+      _protocols[_protocolIds[i]]._referrers.push(_referrerId);
+    }
+    emit ReferrerRegistered(_referrerId, _protocolIds, _rewardRates, _rewardAddress);
+  }
+
+  function registerReferral(
```
silasbw, line 70: So, for the same `msg.sender` both of these could succeed, correct?

```solidity
registerReferral('foo', 'beefy');
registerReferral('goo', 'beefy');
```

What would be the plan for determining revenue attribution then?  Up to the off-chain caller to use the timestamp property?  Would it make sense to revert with an error in instance of the second call (where `msg.sender` is already registered for a protocol).

contracts/Registry.sol
```diff
@@ -0,0 +1,114 @@
...
+    if (
+      // Check if the referrer has been initialized but the user has not been registered before
+      _referrers[protocolId][referrerId].rewardAddress != address(0) &&
+      _users[protocolId][referrerId][msg.sender].timestamp == 0
+    ) {
+      _users[protocolId][referrerId][msg.sender] = User(block.timestamp);
+      _referrers[protocolId][referrerId].userAddresses.push(msg.sender);
+      emit ReferralRegistered(protocolId, referrerId, msg.sender);
+    } else {
+      emit ReferralSkipped(protocolId, referrerId, msg.sender);
```
silasbw, line 83: Without really knowing more about usage, it seems like a revert here is what clients/callers would expect (rather than, essentially, a fail silently).

contracts/Registry.sol
finnian0826: ahh yeah the check I had was wrong, fixed it in [ff625eb](https://github.com/mobilestack-xyz/funding-layer/pull/3/commits/ff625ebbd37d43d83b848224a239044ddcd2163b) to check if a user is registered to a given protocol by any referrer. Working now to have it revert with an error rather than emit `ReferralSkipped`.

finnian0826: custom errors added to revert in [e5abd46](https://github.com/mobilestack-xyz/funding-layer/pull/3/commits/e5abd46f8bdffb5fc12873cf56173ff93c49b425)

contracts/Registry.sol
```diff
@@ -0,0 +1,126 @@
...
+  function registerReferral(
+    string calldata referrerId,
+    string calldata protocolId
+  ) external {
+    // Check if the referrer has been initialized but the user has not been registered before to the given protocol
+    if (_referrers[protocolId][referrerId].rewardAddress == address(0)) {
+      revert ReferrerNotRegistered(protocolId, referrerId);
+    } else if (_usersByProtocol[protocolId][msg.sender].timestamp != 0) {
+      revert UserAlreadyRegistered(protocolId, referrerId, msg.sender);
+    } else {
```
silasbw, line 88: is a single mapping sufficient here?  `_referrer[protocolId][msg.sender] = reffererId`?

contracts/Registry.sol
```diff
@@ -0,0 +1,126 @@
...
+    uint256 rewardRate;
+    address[] userAddresses;
+  }
+
+  struct Protocol {
+    string[] referrers;
+  }
+
+  mapping(string => Protocol) private _protocols;
+  mapping(string => mapping(string => Referrer)) private _referrers;
```
silasbw, line 47: i would name this something else, since the data here doesn't represent a referrer necessarily.  maybe `_partner` or `_frontend`?  There's only a referrer once there's a referral, so it took a second for me to understand the intent of the property was more around potential referrers.

contracts/Registry.sol
finnian0826: What would the single mapping replace? For a given referrer/protocol pair we want to get all the referred users, using the above data structure we'd have to loop over every user for a given protocol and check their referrer which seems inefficient

contracts/Registry.sol
finnian0826: There doesn't need to be a referral for the referrer to exist in `_referrers`, when `registerReferrer` is called, the info for the referrer is added. The data here should be a map from protocolId to a mapping from referrerId to `Referrer`, which is all the info about a specific referrer.

contracts/Registry.sol
finnian0826: The reason I have two is because there are two cases that need to be handled around users:
- For a given protocol, check if a user has been referred to it by ANY referrer
- For a given protocol and a given referrer, get all the users they have referrered

Within the constraints of solidity that I've seen (can't get object.keys, etc), it seemed most efficient to have two different data structures

finnian0826: I don't think the code you added would work, because for a given referrerId we want to get all the users they referred, and that would require us to iterate over every user and check if referrerId matched, but from what I've seen in solidity you can't iterate over keys, if you want to do that you have to store a key array (or something similar).

I was able to clean it up to just use one mapping in [4bf7331](https://github.com/mobilestack-xyz/funding-layer/pull/3/commits/4bf7331d5545e037ccf27ef5de3a4dce4861b4ec) though rather than the 2 different users ones.

contracts/Registry.sol
jophish: I might add brief comments explaining the intent of these pieces of state. For example "`_protocols is a mapping from protocol IDs to lists of referrer IDs that refer to that protocol, used to perform reverse lookups of all referrers that refer to a particular protocol."` and `"_referrers is a mapping from protocol IDs to mappings of referrer IDs to referrer objects.`.

I find that in order to accomplish apparently simple things in Solidity, you often need to maintain seemingly redundant state, and it's often difficult to understand at a glance why certain choices were made around state management. I like to add comments explaining these choices in Solidity.

contracts/Registry.sol
```diff
@@ -0,0 +1,121 @@
...
+  mapping(string => mapping(address => User)) private _usersByProtocol;
+
+  constructor(
+    address owner,
+    uint48 transferDelay
+  ) AccessControlDefaultAdminRules(transferDelay, owner) {}
+
+  function registerReferrer(
```
jophish, line 55: We may want to potentially _update_ referrer information over time, say, to add/remove a protocol from a referrer, or to update the reward rate for a particular protocol that a referrer supports. Calling `registerReferrer` multiple times with the same `_referrerId` but varying the other parameters can potentially leave the contract state inconsistent/incorrect. I think it would be good if `registerReferrer` essentially acted as an "overwrite", meaning it would functionally "wipe" the state of any record of `_referrerId` before running the existing body of this method. This would also give the function the property of being idempotent, which is nice. If we wanted to remove or add a protocol, or update the reward rates or something, we would just call this method with whatever we last called it with, plus whatever diff we want.

This means that we would need some way to retrieve the current information about a particular referrer in order to call `registerReferrer` with the correct parameters whenever we want to perform an update. In particular, for some referrer, we should be able to get:
* The associated protocols
* Those protocol's reward rates
* The reward address

Seeing as how we currently have no mapping in the state whose top-level is keyed by referrer ID, I don't think a getter like this is possible without augmenting the state in some way, since, like you said, we cannot iterate over keys in Solidity.

contracts/Registry.sol
```diff
@@ -0,0 +1,121 @@
...
+    string calldata _referrerId,
+    string[] calldata _protocolIds,
+    uint256[] calldata _rewardRates,
+    address _rewardAddress
+  ) external onlyRole(DEFAULT_ADMIN_ROLE) {
+    for (uint256 i = 0; i < _protocolIds.length; i++) {
+      _referrers[_protocolIds[i]][_referrerId] = Referrer(
```
jophish, line 62: A `Referrer` struct is not actually a "global" object per referrer; it seems to me to be implicitly tied to a particular protocol by virtue of its placement in the `_referrers` mapping. However, it does contain a piece of state, `rewardAddress`, which _is_ global/unique to a particular referrer. With the current state setup, there is no real source of truth for `rewardAddress` for a particular referrer, since it may be duplicated several times among different `Referrer` objects. Kind of a nitpick, but I might suggest moving `rewardAddress` to a mapping from referrer ID to reward address.

Also (as I mention in a previous comment), we currently have no way to retrieve reward address from the contract state for a particular referrer.

jophish (commented):
Looks awesome -- left some comments with suggestions. The state management here is definitely trickier than I had anticipated.

silasbw (approved):
now that's joe's back, i shouldn't be blocking this anymore.

contracts/Registry.sol
```diff
@@ -0,0 +1,161 @@
...
+          keccak256(abi.encodePacked(referrerId))
+        ) {
+          // Replace the current element with the last element
+          referrerIds[j] = referrerIds[referrerIds.length - 1];
+          referrerIds.pop(); // Remove the last element
+          break; // Exit the inner loop after removal
+        }
+      }
+    }
+    // Add/update the reward rate for each provider, add the referrer to the list of referrers for each provider
```
jophish, line 79: nit:
```suggestion
    // Add/update the reward rate for each protocol, add the referrer to the list of referrers for each protocol
```

for consistency

contracts/Registry.sol
```diff
@@ -0,0 +1,161 @@
...
+    } else {
+      _userInfoByProtocol[protocolId][msg.sender] = User(block.timestamp);
+      _referrerInfoByProtocol[protocolId][referrerId].userAddresses.push(
+        msg.sender
+      );
+      emit ReferralRegistered(protocolId, referrerId, msg.sender);
+    }
+  }
+
+  function getReferrers(
```
jophish, line 126: Can we add a corresponding `getProtocols` method to match this?

test/Registry.ts
```diff
@@ -0,0 +1,278 @@
...
+        mockReferrerId2,
+        [mockProtocolId],
+        mockRewardRates,
+        mockRewardAddress,
+      )
+      await registry
+        .connect(addr1)
+        .registerReferral(mockReferrerId, mockProtocolId)
+
+      // Trying to register again should emit "ReferralSkipped"
```
jophish, line 199:
```suggestion
      // Trying to register again should emit "UserAlreadyRegistered"
```

contracts/Registry.sol
```diff
@@ -0,0 +1,161 @@
...
+        }
+      }
+    }
+    // Add/update the reward rate for each provider, add the referrer to the list of referrers for each provider
+    for (uint256 i = 0; i < protocolIds.length; i++) {
+      _referrerInfoByProtocol[protocolIds[i]][referrerId]
+        .rewardRate = rewardRates[i];
+      _protocolIdToReferrerIds[protocolIds[i]].push(referrerId);
+      // Update the list of protocols for the referrer (need to copy each element individually, cannot just assign the array)
+      _referrerIdToProtocolIds[referrerId].push(protocolIds[i]);
```
jophish, line 85: It looks like we never actually reset this list on subsequent calls to `registerReferrer`; we only ever append to it. By my logic, if we call `registerReferrer` for `referrerA` and `protocolA, protocolB`, and then later call it for `referrerA`, `protocolC`, `_referrerIdToProtocolIds[referrerA]` would end up being `[protocolA, protocolB, protocolC]`, no?

I think one consequence of this is that a user may be able to register themselves with a referrer/protocol pair that has since been de-registered (like `referrerA`/`protocolA` above), since the loop on [line 102](https://github.com/mobilestack-xyz/funding-layer/pull/3/files#diff-3579d55f1ea85bbe0e06236275421f84e98afd945520c4138c2c4cb1a7b3543fR102) will set `_referrerIsRegistered` to `true`. 

Can we add a test for this? The contents of the state itself is more directly testable if we have a `getProtocols` getter as well, as mentioned below.

jophish (commented):
Looks super close! One comment about a potential logic bug where we never reset the mapping of `referrerId` to `protocolId`s, which could cause issues, and a suggestion for an additional getter.

jophish (approved):
Looks great!! Thanks for making all the changes. I think you can resolve the knip issue by adding `ethers` as a dev dependency, `yarn add --dev ethers`

PR #5, opened 2024-12-19
feat(contract): add modified jumpstart contract

finnian0826:
Copied over the WalletJumpstart contract, made two modification which are documented in comments below.

### DIFF
--- a/contracts/WalletJumpstartV2.sol
+++ b/contracts/WalletJumpstartV2.sol
@@ -1,0 +1,215 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+import {IERC20} from '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+import {IERC721} from '@openzeppelin/contracts/token/ERC721/IERC721.sol';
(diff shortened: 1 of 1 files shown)


contracts/WalletJumpstartV2.sol
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  struct ERC721Claim {
+    IERC721 token;
+    address depositor;
+    uint256 tokenId;
+    bool claimed;
+  }
+
+  mapping(address => address) public referrer;
```
finnian0826, line 33: this field added

contracts/WalletJumpstartV2.sol
```diff
@@ -0,0 +1,215 @@
...
+    address signer = hash.toEthSignedMessageHash().recover(signature);
+    if (beneficiary != signer) {
+      revert InvalidSignature();
+    }
+
+    claim.claimed = true;
+    claim.token.transfer(sendTo, claim.amount);
+    if (referrer[sendTo] == address(0)) {
+      referrer[sendTo] = claim.depositor;
+    }
```
finnian0826, line 138: This added to map a user to their referrer

PR #7, opened 2025-01-10
feat(revenue): Add initial skeleton revenue script

jophish:
This is the first PR in a chain to get an initial version of the Beefy revenue attribution script working. See this [closed PR](https://github.com/mobilestack-xyz/funding-layer/pull/6#issue-2778995969) for a description/high-level overview of what this will eventually look like; I closed that PR in favor of a series of smaller PRs. I recommend reading the overview linked above to get a sense of what exactly these PRs are working towards.

This PR introduces a skeleton framework for writing handlers for calculating revenue attribution for different protocols.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,6 +8,16 @@ module.exports = {
     'plugin:import/recommended',
     'plugin:import/typescript',
   ],
+  rules: {
+    '@typescript-eslint/no-unused-vars': [
+      'error',
+      {
+        argsIgnorePattern: '^_',
+        varsIgnorePattern: '^_',
+        caughtErrorsIgnorePattern: '^_',
+      },
+    ],
+  },
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint', 'import'],
   settings: {
--- a/scripts/calculateRevenue/calculateRevenue.ts
+++ b/scripts/calculateRevenue/calculateRevenue.ts
@@ -17,11 +17,11 @@ async function main(args: ReturnType<typeof parseArgs>) {
     console.log(
       `Calculating revenue for ${address} (${i + 1}/${eligibleAddresses.length})`,
     )
-    const userResult = await handler(
+    const userResult = await handler({
       address,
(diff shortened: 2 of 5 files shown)


.eslintrc.js
```diff
@@ -8,6 +8,16 @@ module.exports = {
     'plugin:import/recommended',
     'plugin:import/typescript',
   ],
+  rules: {
+    '@typescript-eslint/no-unused-vars': [
+      'error',
+      {
+        argsIgnorePattern: '^_',
+        varsIgnorePattern: '^_',
+        caughtErrorsIgnorePattern: '^_',
```
satish-ravi, line 17: is this temporary for the skeleton function you have below? If so can you just use eslint-disable comment?

scripts/calculateRevenue/calculateRevenue.ts
```diff
@@ -0,0 +1,79 @@
...
+  const handler = calculateRevenueHandlers[args['protocol-id'] as Protocol]
+
+  const allResults: Record<string, RevenueResult> = {}
+
+  for (let i = 0; i < eligibleAddresses.length; i++) {
+    const address = eligibleAddresses[i]
+    console.log(
+      `Calculating revenue for ${address} (${i + 1}/${eligibleAddresses.length})`,
+    )
+    const userResult = await handler(
```
satish-ravi, line 20: is this await better to do it serially? or can we use Promise.all?

silasbw (approved):

.eslintrc.js
jophish: This will be removed once everything's done. For some reason I thought that we already had this rule baked into our config somewhere, but apparently not.

scripts/calculateRevenue/calculateRevenue.ts
jophish: I worry about running into 429s if we parallelize all these calls. Eventually, the helper methods responsible for fetching data will be memoized/cached, so subsequent calls should be much faster.

jh2oman (approved):

PR #8, opened 2025-01-10
feat(revenue): Add Beefy helper to fetch vault TVL

jophish:
Follow-up PR to #7. This PR introduces a helper function that calls a public [Beefy API endpoint](https://databarn.beefy.com/api/v1/documentation/static/index.html) in order to get historical vault TVL (in USD) in 15-minute increments. This will be used later on when calculating a user's share of a vault at any given time, as the denominator in the ratio.

PR #9, opened 2025-01-10
feat(revenue): Add Viem helper to fetch vault fee events

jophish:
Followup PR to #8. This PR adds a Viem helper to fetch the history of fees charged by a vault across a given time range. Later on, we'll use this data by iterating over all the fees charged in the desired time range, and multiplying each by the user's share of the vault's TVL in order to determine how much of the fee is attributable to each user.

In addition to using Viem to index events, this also incidentally integrates with a convenient [DefiLlama endpoint](https://defillama.com/docs/api) that returns the block closest to a given timestamp. Doing this ourselves would be pretty cumbersome.


scripts/calculateRevenue/protocols/beefy.ts
```diff
@@ -1,11 +1,100 @@
...
+  const strategyContract = await getStrategyContract(vaultAddress, networkId)
+
+  const startBlock = await getNearestBlock(networkId, startTimestamp)
+  const endBlock = await getNearestBlock(networkId, endTimestamp)
+  const blocksPer = 10000
+  let currentBlock = startBlock
+
+  const feeEvents: FeeEvent[] = []
+  while (currentBlock < endBlock) {
+    const toBlock = Math.min(currentBlock + blocksPer, endBlock)
```
silasbw, line 79: Does this double-count the toBlock?  Like, imagine "blocksPer" is 1:

currentBlock = 0
toBlock = 1
getEvents({ fromBlock: currentBlock, toBlock }) ---> returns events from 0, and 1

currentBlock = 1
toBlock = 2
getEvents({ fromBlock: currentBlock, toBlock }) ---> returns events from 1, and 2

?

scripts/calculateRevenue/protocols/beefy.ts
```diff
@@ -1,11 +1,100 @@
...
 
+const NETWORK_ID_TO_DEFI_LLAMA_CHAIN: Partial<{
+  [networkId in NetworkId]: string // eslint-disable-line @typescript-eslint/no-unused-vars
+}> = {
+  [NetworkId['ethereum-mainnet']]: 'Ethereum',
+  [NetworkId['arbitrum-one']]: 'Arbitrum',
+  [NetworkId['op-mainnet']]: 'Optimism',
+  [NetworkId['celo-mainnet']]: 'Celo',
+  [NetworkId['polygon-pos-mainnet']]: 'Polygon',
+  [NetworkId['base-mainnet']]: 'base',
```
satish-ravi, line 20: interesting that this is lowercase

scripts/calculateRevenue/protocols/beefy.ts
```diff
@@ -1,11 +1,100 @@
...
+): Promise<number> {
+  const unixTimestamp = Math.floor(timestamp.getTime() / 1000)
+  const defiLlamaChain = NETWORK_ID_TO_DEFI_LLAMA_CHAIN[networkId]
+
+  const response = await fetchWithTimeout(
+    `${DEFI_LLAMA_API_URL}/block/${defiLlamaChain}/${unixTimestamp}`,
+  )
+  if (!response.ok) {
+    console.log(
+      `${DEFI_LLAMA_API_URL}/block/${defiLlamaChain}/${unixTimestamp}`,
```
satish-ravi, line 49: nit: pull this to a variable, so these don't accidentally diverge

scripts/calculateRevenue/protocols/beefy.ts
jophish: Wow, I just assumed that the end range would be exclusive, but in fact it appears that both limits are _inclusive_, according to [here](https://ethereum.stackexchange.com/questions/78620/is-the-range-of-a-filterquery-inclusive-up-to-toblock). That's kind of annoying -- thanks for the callout.

scripts/calculateRevenue/protocols/beefy.ts
jophish: Oops, this log actually shouldn't have made it in to the PR.

scripts/calculateRevenue/protocols/beefy.ts
jophish: Actually, I think these can all be lowercase. The endpoint accepts certain chains as both uppercase/lowercase, but base only worked when lowercase. After some testing, it looks like all chains work when lowercased, so I'll just do that for them all.

scripts/abis/BeefyVaultV7.ts
MuckT, line 1: A comment with a URL for where this ABI was sourced from might come in handy in the future.

scripts/abis/StratFeeManagerInitializable.ts
MuckT, line 1: A comment on the source of this ABI would be helpful.

scripts/calculateRevenue/protocols/beefy.ts
```diff
@@ -1,11 +1,100 @@
...
+// TODO: Memoize this function so it's not repeated for every user address
+/**
+ * For a given vault, fetches the record of all ChargedFee events emitted in a given timeframe
+ */
+export async function fetchFeeEvents(
+  vaultAddress: Address,
+  networkId: NetworkId,
+  startTimestamp: Date,
+  endTimestamp: Date,
+): Promise<FeeEvent[]> {
```
satish-ravi, line 66: use object params especially if 2 types are the same. I think we should favor object params in general, maybe positional is fine for 1 or max 2 params if they are of different types

scripts/calculateRevenue/protocols/beefy.ts
```diff
@@ -1,11 +1,100 @@
...
+  const client = getViemPublicClient(networkId)
+  const strategyContract = await getStrategyContract(vaultAddress, networkId)
+
+  const startBlock = await getNearestBlock(networkId, startTimestamp)
+  const endBlock = await getNearestBlock(networkId, endTimestamp)
+  const blocksPer = 10000
+  let currentBlock = startBlock
+
+  const feeEvents: FeeEvent[] = []
+  while (currentBlock < endBlock) {
```
satish-ravi, line 76: should this be `<=`? Otherwise I don't see the point of the Math.min in the last line of the loop

jophish: Ah, you're right. I think it's simpler to keep the `<` and just do `currentBlock = toBlock + 1`

jh2oman (approved):

PR #10, opened 2025-01-10
feat(revenue): Plumb revenue script helpers, get initial version working

jophish:
Followup PR to #9. This PR plumbs together all of the helpers from the previous PRs, and introduces some new logic to actually perform fee attribution on the data.

To facilitate easier testing (since internal helper functions cannot be mocked by jest), I've also refactored the organization of some of the files/functions. Git appears to not be able to recognize the moved files as being the same as before, so the actual diff size in this PR is smaller than what's reported. I'll add some comments to files that can be safely ignored.

There are some follow-ons and caveats from the [original PR](https://github.com/mobilestack-xyz/funding-layer/pull/6#issue-2778995969) that apply here.


scripts/calculateRevenue/protocols/beefy/helpers.ts
jophish, line 1: This file is just a port of the original `beefy.ts` contents - you can ignore this.

scripts/calculateRevenue/protocols/beefy/types.ts
```diff
@@ -0,0 +1,30 @@
+import { BeefyInvestorTransaction } from '../../../protocol-filters/beefy'
```
jophish, line 1: You can ignore the existing types here, I just pulled them out from the old `beefy.ts` since they're now shared more widely.

scripts/calculateRevenue/protocols/beefy/beefy.ts
satish-ravi, line 1: nit: can be named index.ts? To avoid `beefy/beefy`

scripts/calculateRevenue/protocols/beefy/beefy.ts
```diff
@@ -0,0 +1,112 @@
...
+  BeefyInvestorTransactionWithUsdBalance,
+  BeefyVaultTvlData,
+  VaultInfo,
+} from './types'
+import { getStrategyContract } from '../utils/viem'
+import { RevenueResult, NetworkId } from '../../../types'
+import { getVaults } from './getVaults'
+
+/**
+ * Given a list of transaction history on a particular Beefy vault and a timestamp,
```
satish-ravi, line 11: nit: make a note that this requires the data to be sorted?

scripts/calculateRevenue/protocols/beefy/beefy.ts
```diff
@@ -0,0 +1,112 @@
...
+    if (new Date(tx.datetime) >= timestamp) {
+      return lastTvl
+    }
+    lastTvl = tx.usd_balance
+  }
+  return lastTvl
+}
+
+/**
+ * Given a list of Beefy vault TVL data and a timestamp, returns the vault's TVL at
```
satish-ravi, line 29: same as above

scripts/calculateRevenue/protocols/beefy/beefy.ts
```diff
@@ -0,0 +1,112 @@
...
+  tokenId: string
+  revenue: string
+}> {
+  let totalRevenueContribution = 0n
+  for (const { beefyFee, timestamp } of vaultInfo.feeEvents) {
+    const userTvl = getUserTvl(vaultInfo.txHistory, timestamp)
+    const vaultTvl = getVaultTvl(vaultInfo.vaultTvlHistory, timestamp)
+    const partialRevenueContribution =
+      (BigInt(userTvl * 10 ** 18) * BigInt(beefyFee)) /
+      BigInt(vaultTvl * 10 ** 18)
```
satish-ravi, line 61: nit: should we do the 10 ** 18 multiplication after converting to bigint to be safe?

satish-ravi (commented):
Looks good, mostly nits

scripts/calculateRevenue/protocols/beefy/getVaults.ts
```diff
@@ -0,0 +1,74 @@
+import { fetchInvestorTimeline } from '../../../protocol-filters/beefy'
```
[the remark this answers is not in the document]
jophish: My preference is camel case, and it's what we typically tend to do. What do you think?

scripts/calculateRevenue/protocols/beefy/getVaults.ts
```diff
@@ -0,0 +1,74 @@
...
+  const vaultsInfo: VaultsInfo = {}
+
+  for (const vault of allVaults) {
+    // Ensure that per-vault transaction history is sorted oldest to newest.
+    const txHistory = portfolioData
+      .filter((data) => data.product_key === vault)
+      .sort((a, b) => {
+        if (new Date(a.datetime) < new Date(b.datetime)) return -1
+        if (new Date(a.datetime) > new Date(b.datetime)) return 1
+        return 0
```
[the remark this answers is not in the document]
jophish: They aren't ISO strings unfortunately, and `return new Date(a.datetime) - new Date(b.datetime)` doesn't work on the Date objects themselves (you have to do something like `return new Date(a.datetime).getTime() - new Date(b.datetime).getTime()` in order for TS not to complain).

scripts/calculateRevenue/protocols/beefy/beefy.ts
jophish: We have to do the conversion before; converting the values to BigInt directly doesn't work since they have lots of decimal precision, and rounding would lose all of that.

scripts/calculateRevenue/protocols/beefy/getVaults.ts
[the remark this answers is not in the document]
jophish: although, for the actual script names, I suppose it's more idiomatic to use kebab case.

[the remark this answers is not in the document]
satish-ravi: I think being consistent one way or the other would be nice. Don't really have a preference, we can just stick to camel case everywhere if that's the easiest.

satish-ravi (approved):

scripts/calculateRevenue/protocols/beefy/getVaults.ts
```diff
@@ -0,0 +1,72 @@
...
+    (vaults, data) => vaults.add(data.product_key),
+    new Set<string>(),
+  )
+
+  const vaultsInfo: VaultsInfo = {}
+
+  for (const vault of allVaults) {
+    // Ensure that per-vault transaction history is sorted oldest to newest.
+    const txHistory = portfolioData
+      .filter((data) => data.product_key === vault)
```
MuckT, line 45: Instead of filtering here can we instead group all vaults by product key before?

PR #13, opened 2025-01-15
feat(referral): fetch referrals from registry contract

jh2oman:
* fetches referrals from all networks (because if two different frontends both refer a user to a protocol on different chains we need to determine which one did it first)
* removes duplicate referals

### DIFF
--- a/scripts/fetch-referrals.ts
+++ b/scripts/fetch-referrals.ts
@@ -28,7 +28,6 @@ async function getArgs() {
   return {
     protocol: argv['protocol'] as Protocol,
     protocolFilter: protocolFilters[argv['protocol'] as Protocol],
-    input: argv['input'],
     output: argv['output'],
   }
 }
@@ -50,7 +49,7 @@ async function main() {
 
   const filteredEvents = await args.protocolFilter(uniqueEvents)
   const output = filteredEvents
-    .map((event) => `${event.userAddress},${event.timestamp}`)
+    .map((event) => `${args.protocol},${event.referrerId},${event.userAddress},${event.timestamp}`)
     .join('\n')
   writeFileSync(args.output, output)
 }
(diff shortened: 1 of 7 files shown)


finnian0826 (commented):
It seems like the output only contains the filtered referred users and their timestamp, not the address of the user that referred them. How are we planning on using this output file? Without the referring user in the output file or passed in as a param I'm a bit confused

jh2oman:
@finnian0826 oh! Excelent point!

scripts/fetch-referrals.ts
```diff
@@ -19,42 +18,38 @@ async function getArgs() {
...
+    NetworkId['base-mainnet'],
+  ]
+
+  const referralEvents = await fetchReferralEvents(networkIds, args.protocol)
+  const uniqueEvents = removeDuplicates(referralEvents)
+
+  const filteredEvents = await args.protocolFilter(uniqueEvents)
   const output = filteredEvents
-    .map((event) => `${event.userAddress},${event.timestamp}`)
+    .map((event) => `${args.protocol},${event.referrerId},${event.userAddress},${event.timestamp}`)
```
finnian0826, line 52: including protocol might be redundant, since a given file is only for one protocol so the file could be named `beefy_filtered_referrals.csv` rather than having a `Beefy` entry on every line

finnian0826 (approved):
small feedback on output file but otherwise lgtm!

scripts/utils.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (!client) {
+    throw new Error(`No viem client found for networkId: ${networkId}`)
+  }
+  return client
+}
+
+/**
+ * Returns a contract object representing the registry
+ */
+export async function getRegistryContract(
```
jophish, line 53: nit: This could just accept a `networkId` and do the translation internally, since there's only going to be one registry per network.

jophish (commented):
A little late to the party, but this looks great to me!

PR #14, opened 2025-01-15
feat(revenue): Add token price history to normalize fees, simplify output format

jophish:
Followup to #10. Adds in historical token price fetching in order to normalize fees paid to a common unit. Also simplifies the output format to just be a simple CSV, since there's no obvious use case atm for more detailed data.


scripts/calculateRevenue/protocols/beefy/index.ts
```diff
@@ -16,7 +20,7 @@ export function getUserTvl(
   txHistory: BeefyInvestorTransactionWithUsdBalance[],
   timestamp: Date,
 ): number {
-  let lastTvl = txHistory[0].usd_balance
+  let lastTvl = 0
```
jophish, line 23: This is a small bugfix. Since we're assuming that the Beefy user TX data is complete, we should be assuming that the user starts out with 0 balance in all vaults.

scripts/abis/ERC20.ts
```diff
@@ -0,0 +1,224 @@
+const ERC20 = [
```
satish-ravi, line 1: could possibly get this from viem too

scripts/calculateRevenue/protocols/beefy/index.ts
```diff
@@ -45,37 +49,77 @@ export function getVaultTvl(
   return lastTvl
 }
 
+/**
+ * Given a list of historical token prices and a timestamp, returns the nearest token price
+ * to the specified time.
+ * Requires the list of tokenPrices to be sorted from earliest to most-recent.
+ */
+export function getTokenPrice(
```
satish-ravi, line 57: how accurate does this need to be? Since we're using the first price after the given timestamp, if there's a large gap in the data, we might be off by a bit.

scripts/calculateRevenue/protocols/beefy/index.ts
```diff
@@ -4,8 +4,12 @@ import {
   VaultInfo,
 } from './types'
 import { getStrategyContract } from '../utils/viem'
-import { RevenueResult, NetworkId } from '../../../types'
+import { TokenPriceData } from '../../../types'
 import { getVaults } from './getVaults'
+import { fetchTokenPrices } from '../utils/tokenPrices'
+import { getErc20Contract } from '../../../utils'
+
+const REVENUE_USD_PRECISION = 8
```
satish-ravi, line 12: why 8?

scripts/calculateRevenue/protocols/beefy/index.ts
```diff
@@ -45,37 +49,77 @@ export function getVaultTvl(
...
+  vaultInfo: VaultInfo,
+): Promise<number> {
   const strategyContract = await getStrategyContract(
     vaultInfo.vaultAddress,
     vaultInfo.networkId,
   )
   const nativeTokenAddress = await strategyContract.read.native()
-  const tokenId = `${vaultInfo.networkId}:${nativeTokenAddress}`
-  return {
+  const tokenId = `${vaultInfo.networkId}:${nativeTokenAddress.toLowerCase()}`
```
satish-ravi, line 83: aren't native tokens `<network-id>:native`?

scripts/calculateRevenue/calculateRevenue.ts
MuckT, line 1: [p-map](https://github.com/sindresorhus/p-map) might be a good solution if we need to calculate these revenues often, but it is also something we can add later if needed.

scripts/abis/ERC20.ts
jophish: How? Does Viem export this? I can't seem to find it anywhere.

scripts/calculateRevenue/protocols/beefy/index.ts
jophish: I believe our endpoint (at least as far as I saw) returns hourly prices. It's not perfect granularity, but I think it will probably suffice, especially for this v0.

scripts/calculateRevenue/protocols/beefy/index.ts
jophish: Somewhat arbitrary, just needed good-enough precision since revenue will often be fractions of dollars. Added a comment.

scripts/calculateRevenue/protocols/beefy/index.ts
jophish: Ah good call. For chains where the native token is something like wrapped ETH, we can use the address, but `native` is more universal.

scripts/abis/ERC20.ts
jh2oman: I feel like it *should* but also can't find it either 😞

scripts/calculateRevenue/protocols/beefy/index.ts
```diff
@@ -45,37 +51,77 @@ export function getVaultTvl(
...
+
+  for (const { beefyFee, timestamp } of vaultInfo.feeEvents) {
+    const userTvl = getUserTvl(vaultInfo.txHistory, timestamp)
+    const vaultTvl = getVaultTvl(vaultInfo.vaultTvlHistory, timestamp)
+
+    const tokenPriceUsd = getTokenPrice(tokenPrices, new Date(timestamp))
+
+    const partialNativeContribution =
+      (BigInt(userTvl * 10 ** 18) * BigInt(beefyFee)) /
+      BigInt(vaultTvl * 10 ** 18)
```
jh2oman, line 112: Don't the `10 ** 18`s cancel each other out?

scripts/calculateRevenue/protocols/beefy/index.test.ts
```diff
@@ -140,17 +162,21 @@ describe('Beefy revenue calculation', () => {
...
-      const expected = {
-        tokenId: mockNativeTokenId,
-        revenue: '1100',
-      }
-      expect(result).toEqual(expected)
+      // native token has 4 decimals, so 1 token is 10000 wei
+      // First fee of 1000 occurs while vault has 1000, user has 100 -> 100/1000 * 1000 = 100 wei
+      // - token price is 1000 -> 100 / 10000 * 1000 USD = 10 USD
+      // Second fee of 5000 occurs while vault has 2000, user has 400 -> 400/2000 * 5000 = 1000 wei
+      // - token price is 1500 -> 1000 / 10000 * 1500 USD = 150 USD
```
jh2oman, line 178: Immensely helpful 🙏

jh2oman (approved):

scripts/calculateRevenue/calculateRevenue.ts
```diff
@@ -10,22 +10,31 @@ async function main(args: ReturnType<typeof parseArgs>) {
...
+      address,
+      revenue,
+    })
   }
+  const output = allResults
+    .map((result) => `${result.address},${result.revenue}`)
+    .join('\n')
 
-  writeFileSync(args['output-file'], JSON.stringify(allResults, null, 4))
+  writeFileSync(args['output-file'], output)
```
MuckT, line 37: Shouldn't matter, but to be safe.
```suggestion
  writeFileSync(args['output-file'], output, { encoding: 'utf-8' })
```

PR #15, opened 2025-01-16
feat(revenue): Memoize functions that are likely to be re-used across users

jophish:
This PR adds in some basic memoization to functions that are likely to be called with the same parameters across users; this is basically a requirement in order to get this to run for many users in a reasonable amount of time, since we pull a _lot_ of data to cobble these results together. In particular, the functions being memoized, and a rationale for their memoization, are given below.

* `fetchVaultTvlHistory`
  * For a given vault, this will _always_ be called with the same parameters across users in that vault (start/end timestamps are fixed).
* `getNearestBlock`
  * This is only ever called on the start/end timestamps given as parameters to the script.
* `fetchFeeEvents`
  * For a given vault, this will _always_ be called with the same parameters across users in that vault (start/end timestamps are fixed).
* `fetchTokenPrices`
  * That this will be called with identical parameters between users is less obvious than the other functions; for a given vault, it is called with the earliest/latest timestamps among fee events for that vault; these fee event timestamps do not change across users of the same vault.

As a dumb example, I told the script to compute results for the same user address 4 times and added some logging:

```bash
$ /home/joe/Documents/valora/funding-layer/node_modules/.bin/ts-node scripts/calculateRevenue/calculateRevenue.ts -i input2.csv -o out.csv -p Beefy -s 1735788429000 -e 1736980031000
Calculating revenue for 0xD6A8ADB4B2EFA66C44568ddeD769F65d86e65f1F (1/4)
handler: 1:06.756 (m:ss.mmm)
Calculating revenue for 0xD6A8ADB4B2EFA66C44568ddeD769F65d86e65f1F (2/4)
handler: 512.756ms
Calculating revenue for 0xD6A8ADB4B2EFA66C44568ddeD769F65d86e65f1F (3/4)
handler: 492.549ms
Calculating revenue for 0xD6A8ADB4B2EFA66C44568ddeD769F65d86e65f1F (4/4)
handler: 714.55ms
Wrote results to out.csv
```


scripts/calculateRevenue/protocols/beefy/helpers.ts
```diff
@@ -44,11 +41,14 @@ export async function getNearestBlock(
   return blockTimestampData.height
 }
 
-// TODO: Memoize this function so it's not repeated for every user address
+export const getNearestBlock = memoize(_getNearestBlock, {
+  hash: (...params: Parameters<typeof _getNearestBlock>) => params.join(','),
```
satish-ravi, line 45: what's the default hash function?

satish-ravi (approved):
Nice!

scripts/calculateRevenue/protocols/beefy/helpers.ts
MuckT: This is what I found: https://github.com/github/memoize/blob/main/index.ts#L2-L6.

PR #18, opened 2025-01-23
chore: tweak argument handling in deploy script

silasbw:


jophish (approved):

scripts/deploy.ts
```diff
@@ -16,15 +16,27 @@ async function getConfig() {
...
       description: 'Salt to use for CREATE2 deployments',
       type: 'string',
-      demandOption: true,
     })
     .option('owner-address', {
       description: 'Address of the address to use as owner',
       type: 'string',
       demandOption: true,
+    })
+    .check((argv) => {
```
jophish, line 33: Might be nice to ensure (below) that we only actually use Defender if `use-defender` is true.

scripts/deploy.ts
```diff
@@ -16,15 +16,27 @@ async function getConfig() {
...
+      if (argv.useDefender && !SUPPORTED_NETWORKS.includes(hre.network.name)) {
```
MuckT, line 34: If I understand yargs correctly, if a `deploy-salt` is not provided, `argv.useDefender` will be false, and this check won't run. Is that correct? If so, LGTM.

MuckT (approved):

scripts/deploy.ts
silasbw: oops -- good catch

PR #20, opened 2025-01-24
feat: add referrer user count script

MuckT:
### Description

Part of [ACT-1493](https://linear.app/valora/issue/ACT-1493/write-script-to-collect-informative-metrics) adds a script that returns the number of eligible users referred to a protocol by a referrer. The script is pretty granular, requiring a referrer address(es) and outputting a file for each. If we want a script to return the user count by all referrers for a protocol, then a new script / additional script will be needed.


scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,74 @@
...
+async function getArgs() {
+  const argv = await yargs
+    .env('')
+    .option('protocol', {
+      description: 'protocol that the referrals are for',
+      demandOption: true,
+      choices: protocols,
+    })
+    // This is coerced as a string array to prevent 0x... from being interpreted as a number
+    .option('referrerAddresses', {
```
silasbw, line 16: if you name it `foo-bar`, yargs automatically provides two properties: `['foo-bar'] and ['fooBar']`.  So use the dashed naming.

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,74 @@
+import yargs from 'yargs'
+import { protocolFilters, supportedNetworkIds } from './consts'
+import { fetchReferralEvents, removeDuplicates } from './referrals'
+import { NetworkId, Protocol, protocols } from './types'
+import { writeFileSync } from 'fs'
+
+async function getArgs() {
+  const argv = await yargs
+    .env('')
+    .option('protocol', {
```
silasbw, line 10: add a type

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,74 @@
...
+      type: 'string',
+      demandOption: true,
+      coerce: (arg) => new Set(arg.split(',')),
+    })
+    .option('networkIds', {
+      description: 'Comma-separated list of network IDs',
+      type: 'string',
+      demandOption: false,
+      coerce: (arg: string) => {
+        const networkIds = arg.split(',').map((id) => id.trim())
```
silasbw, line 27: why `trim` here, but not for `referrer-addresses`?  i would drop it.

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,74 @@
...
+        const invalidIds = networkIds.filter(
```
silasbw, line 28: This is going to make it wierd to use the local testnet.  Is this check really necessary?

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,74 @@
...
+  )
+  const uniqueEvents = removeDuplicates(referralEvents)
+  const protocolFilteredEvents = await args.protocolFilter(uniqueEvents)
+
+  referrerArray.forEach((referrer) => {
+    const targetReferrerEvents = protocolFilteredEvents.filter(
+      (event) => event.referrerId === referrer,
+    )
+    const output = `${args.protocol},${targetReferrerEvents.length}`
+    writeFileSync(`${referrer}_referral_count.csv`, output, {
```
silasbw, line 65: use a csv write.  we've used https://csv.js.org/ before

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const referralEvents = await fetchReferralEvents(
+    networkIdsArray,
+    args.protocol,
+    referrerArray,
+  )
+  const uniqueEvents = removeDuplicates(referralEvents)
+  const protocolFilteredEvents = await args.protocolFilter(uniqueEvents)
+
+  referrerArray.forEach((referrer) => {
```
silasbw, line 60: i would have this script process a single referrer (which would also simplify your yargs above a little bit).  if we want some giant uber script that does every referrer or something, we can re-use functions from this file.

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,74 @@
...
+    const targetReferrerEvents = protocolFilteredEvents.filter(
+      (event) => event.referrerId === referrer,
+    )
+    const output = `${args.protocol},${targetReferrerEvents.length}`
```
silasbw, line 64: make the output file an argument.

scripts/referrer-user-count.ts
MuckT: We already use `writeFileSync` in `fetch-referrals.ts` and `calculateRevenue.ts`.

silasbw: OK -- we can address those in a sperate PR.

scripts/referrer-user-count.ts
MuckT: @silasbw This user error could occur when someone types `celo-mainet` instead of `celo-mainnet`. With this error, the script chugs along fine, and the output file now omits the user's reference to `celo-mainnet`. I see two solutions for this:
- Leave it as is and create a middleware to validate network IDs used in CLI in a later PR.
- Remove the `networkIds` parameter entirely and the ability to query per network.

scripts/referrer-user-count.ts
MuckT: Added in f0073f9d16b11b770e9eb36dcad8008eb732797d.

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,105 @@
...
+      type: 'string',
+    })
+    .option('output-file', {
+      alias: 'o',
+      description: 'output file path to write JSON results',
+      type: 'string',
+      demandOption: true,
+    })
+    // This is coerced as a string array to prevent 0x... from being interpreted as a number
+    .option('referrer-addresses', {
```
jophish, line 23: A referrer is not an address, but a unique referrer identifier (just an arbitrary string). For example `"valora"`, or a hypothetical frontend to Beefy, `"leefy"`, or something of the sort.

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,105 @@
...
+  const output = results
+    .map((result) => `${result.referrer},${result.referralCount}`)
+    .join('\n')
+  writeFileSync(outputPath, output, { encoding: 'utf-8' })
+}
+
+async function main() {
+  const args = await getArgs()
+  // Conversions to allow yargs to take in a list of referrer addresses / network IDs
+  const referrerArray = args.referrerAddresses
```
jophish, line 58: nit: can we just name these e.g. `referrers` and `networkIds`?

scripts/referrals.ts
```diff
@@ -27,10 +17,11 @@ export function removeDuplicates(events: ReferralEvent[]): ReferralEvent[] {
   return Array.from(uniqueEventsMap.values())
 }
 
-// Fetch all referral events on all networks for the given protocol
+// Fetch referral events on networks for the given protocol with an optional list of referrer addresses
 export async function fetchReferralEvents(
   networkIds: NetworkId[],
   protocol: Protocol,
+  referrerAddresses?: string[],
```
jophish, line 24:
```suggestion
  referrerIds?: string[],
```
And same with the comment above

scripts/referrer-user-count.ts
jophish: I like this pattern fwiw (default to all, optionally specify any subset). The implementation of it is pretty clean imo.

scripts/consts.ts
```diff
@@ -0,0 +1,24 @@
+import { Address } from 'viem'
+import { FilterFunction, NetworkId, Protocol } from './types'
+import { filterEvents as beefy } from './protocol-filters/beefy'
+
+export const NETWORK_ID_TO_REGISTRY_ADDRESS = {
```
jophish, line 5: Since the registry address will always be the same across chains, can we simplify this to just be a single const?

scripts/consts.ts
```diff
@@ -0,0 +1,24 @@
+import { Address } from 'viem'
+import { FilterFunction, NetworkId, Protocol } from './types'
+import { filterEvents as beefy } from './protocol-filters/beefy'
```
jophish, line 3:
```suggestion
import { filterEvents as filterBeefyEvents } from './protocol-filters/beefy'
```

README.md
```diff
@@ -19,7 +19,14 @@ yarn test
...
 ```bash
-npx ts-node ./scripts/fetch-referrals.ts --protocol Beefy --output output.csv
+yarn ts-node ./scripts/fetch-referrals.ts --protocol Beefy --output output.csv
+```
+
+### Fetch Referrals count per referrer
+
+```bash
+# networkIds is optional
+yarn ts-node ./scripts/referrer-user-count.ts --protocol Beefy --referrerAddresses 0x1,0x2,0x3 --networkIds celo-mainnet,base-mainnet
```
jophish, line 29:
```suggestion
yarn ts-node ./scripts/referrer-user-count.ts --protocol Beefy --referrerIds app1,app2,app3 --networkIds celo-mainnet,base-mainnet
```

or something like that

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,105 @@
...
+  )
+  const uniqueEvents = removeDuplicates(referralEvents)
+  const protocolFilteredEvents = await args.protocolFilter(uniqueEvents)
+
+  const allResults: Array<{
+    referrer: string
+    referralCount: number
+  }> = []
+
+  // If referrer addresses are provided output all the addresses and their count
```
jophish, line 78:
```suggestion
  // If referrer IDs are provided output all the IDs and their count
```

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,105 @@
...
+  // Else output all referrer addresses and their count
+  if (referrerArray) {
```
jophish, line 80: `referralEvents` should already only include events for referrer IDs that we actually want to return, right? That is, if we specify some specific referrers, it will only contain referrals from those referrers, else it will have all referrers (the default). We shouldn't need to do anything differently depending on whether or not we've specified specific referrers, right?

jophish, line 80:
```suggestion
  const allResults = protocolFilteredEvents.reduce((acc, event) => {
    const referrer = event.referrerId
    if (acc?.[referrer]) {
      acc.referrer += 1
    } else {
      acc.referrer = 1
    }
    return acc
  }, {})
```
Something like this should be able to sum all the events per referrer, no?

jophish: As it stands now, the two conditions of the if/else are doing logically different things -- the first is summing the event count per referrer (adding a single array entry per referrer) but the second is making a single array entry _per event_.

scripts/consts.ts
MuckT: I split out these changes into https://github.com/mobilestack-xyz/funding-layer/pull/21. If we deploy to ETH mainnet at a later date will it still have the same address?

jophish: Yep, any chains we deploy to in the future will have the same address.

scripts/referrer-user-count.ts
MuckT: Addressed in 7c2b73f6c37b2b44cb7838c79861603c4ecf4ea2!

MuckT: Fixed in 7c2b73f6c37b2b44cb7838c79861603c4ecf4ea2!

scripts/referrer-user-count.ts
MuckT: Opting for option #1 for now.

scripts/consts.ts
MuckT: This object is used in a few other places, so I'm opting to leave it as is and fix it in a subsequent PR.

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,100 @@
...
+      type: 'string',
+      demandOption: false,
+      coerce: (arg: string) => new Set(arg.split(',')),
+    }).argv
+
+  return {
+    protocol: argv['protocol'] as Protocol,
+    protocolFilter: protocolFilters[argv['protocol'] as Protocol],
+    networkIds: (argv['network-ids'] as NetworkId) ?? supportedNetworkIds,
+    referrers: argv['referrer-ids'] as string,
```
silasbw, line 44: isn't this a set?

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,100 @@
...
+    args.protocol,
+    referrerArray,
+  )
+  const uniqueEvents = removeDuplicates(referralEvents)
+  const protocolFilteredEvents = await args.protocolFilter(uniqueEvents)
+
+  const allResultsObj = protocolFilteredEvents.reduce(
+    (acc: Record<string, number>, event) => {
+      const referrer = event.referrerId
+      if (acc?.[referrer]) {
```
silasbw, line 68: nit: `if (referrer in acc)`, which is exactly the condition you want to check for and slightly less error prone than what's written.

silasbw (approved):

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,100 @@
...
+      } else {
+        acc[referrer] = 1
+      }
+      return acc
+    },
+    {},
+  )
+
+  // Handle cases were a cli passed referrer ID param has no results
+  if (referrerArray) {
```
silasbw, line 79: meh, i'd do this first and unconditionally (if there's no referrerArray, just fetch the set of ids first).  It cleans up the logic I comment on above.

scripts/referrer-user-count.ts
MuckT: I had done this to allow yargs to take arrays as `--referrer-ids val1,val2,val3` instead of `--referrer-ids val1 val2 val3`, but I am fine with either option. If we choose to use yargs array type we can cast these to `string[]`.

MuckT: Changed to arrays in dd0514e966c49cd9cd7c7a473e37e00f393c797e.

jophish (approved):

README.md
```diff
@@ -33,7 +33,14 @@ SHELL=true yarn --silent hardhat --network hardhat run scripts/deploy.ts
 Fetch referrals for a specific protocol, removes duplicate events across chains, and filters out events where the user was previously exposed to the protocol
 
 ```bash
-npx ts-node ./scripts/fetch-referrals.ts --protocol Beefy --output output.csv
+yarn ts-node ./scripts/fetch-referrals.ts --protocol Beefy --output output.csv
+```
+
+### Fetch Referrals count per referrer
```
jophish, line 39: Add a summary here describing what this script does?

scripts/referrer-user-count.ts
```diff
@@ -0,0 +1,92 @@
+import { stringify } from 'csv-stringify/sync'
```
jophish, line 1: Can this file be camelCased? That's the convention we've been going with for files/script names.

PR #21, opened 2025-01-24
chore: clean-up types and constants

MuckT:
### Description

Broken out from #20.

- Moves shared types used in scripts to `scripts/types.ts`.
- Moves shared constants used in scripts to `scripts/filters.ts` & `scripts/networks.ts`.
- ~~Updates the funding-layer v0 contract address.~~


scripts/consts.ts
```diff
@@ -0,0 +1,24 @@
+import { Address } from 'viem'
```
silasbw, line 1: can you think of a better filename than "consts"?  How about "networks" or something?

scripts/referrals.test.ts
```diff
@@ -111,11 +111,11 @@ describe('fetchReferralEvents', () => {
     ])
     expect(mockGetRegistryContract).toHaveBeenCalledTimes(2)
     expect(mockGetRegistryContract).toHaveBeenCalledWith(
-      '0x0',
+      '0x5a1a1027aC1d828E7415AF7d797FBA2B0cDD5575',
```
silasbw, line 114: kinda weird to use actual production addresses here.  feels unsafe.  is there an easy way to avoid it?

scripts/consts.ts
MuckT: Changed to `networks.ts` and `filters.ts` in af01c0a2eb8b761c9588f99d9bc0cad8989f2178.

scripts/referrals.test.ts
MuckT: Removed the contract address in af01c0a2eb8b761c9588f99d9bc0cad8989f2178.

silasbw (approved):

PR #22, opened 2025-01-24
chore: "shell" mode for deploy script

silasbw:
This makes it easier to pipe the environment around. E.g.,

```
eval `yarn hardhat run scripts/deploy.ts`
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,20 @@ yarn install
 yarn test
 ```
 
+## Local testnet
+
+Run the localtest in one terminal:
+
+```
+yarn hardhat node
+```
+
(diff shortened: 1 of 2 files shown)


MuckT (commented):
If we're using `shell: true | false` as an indicator of local development, would something like `dev|staging|prod` OR `local|testnet|mainnet` be more appropriate in case we need another value later?

silasbw:
`--shell` or `-s` is the canonical argument in GNU/BSD for this behavior.  e.g., `ssh-agent -s`

MuckT (approved):

PR #23, opened 2025-01-24
chore: update calculateRevenue.ts to use a csv parser

silasbw:


jophish (approved):
this is way nicer -- thanks!

PR #25, opened 2025-01-24
chore: make --owner-address optional

silasbw:


README.md
```diff
@@ -23,7 +23,7 @@ yarn hardhat node
 And deploy in another:
 
 ```
-SHELL=true OWNER_ADDRESS=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266 yarn --silent hardhat --network hardhat run scripts/deploy.ts
+SHELL=true yarn --silent hardhat --network hardhat run scripts/deploy.ts
```
jophish, line 26: Perhaps add a note saying that this will set the owner to the default signer?

PR #28, opened 2025-01-27
feat: add referrer revenue summary script

MuckT:
### Description

Part of [ACT-1493](https://linear.app/valora/issue/ACT-1493) this PR adds a script to get average revenue, total revenue for select or all referrers per protocol for a given time period, defaults which defaults to the lifetime of the protocol.


scripts/referrer-revenue-summary.ts
```diff
@@ -0,0 +1,161 @@
...
+  // Calculate average revenue done here to avoid calculating at each iteration
+  for (const [_, referrerInfo] of referrerMap) {
+    referrerInfo.averageRevenue =
+      referrerInfo.totalRevenue / referrerInfo.referralCount || 0
+  }
+
+  const allResultsArray: [string, number, number, number | undefined][] = []
+
+  // Handle cases were a cli passed referrer ID param that has no results
+  if (referrerArray) {
```
silasbw, line 124: why don't you do this first and unconditionally (if there's not a list of referrers passed in, get the list).  It will clean up the condition about around "if the thing doesn't exist yet, set it to 1"

MuckT: Changed in 08e3adb6bd5e5bc65908f80b9fdce134bdd4979c!

README.md
```diff
@@ -43,6 +43,13 @@ yarn ts-node ./scripts/fetch-referrals.ts --protocol Beefy --output output.csv
 yarn ts-node ./scripts/referrer-user-count.ts --protocol Beefy --referrerIds app1,app2,app3 --networkIds celo-mainnet,base-mainnet
 ```
 
+### Fetch Revenue Summary
```
jophish, line 46: Add a short summary describing what this script actually does/the data is returns?

scripts/referrer-revenue-summary.ts
```diff
@@ -0,0 +1,156 @@
+import yargs from 'yargs'
```
jophish, line 1: Camel case this file?

scripts/referrer-revenue-summary.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+  }
+
+  // Calculate average revenue done here to avoid calculating at each iteration
+  for (const [_, referrerInfo] of referrerMap) {
+    referrerInfo.averageRevenue =
+      referrerInfo.totalRevenue / referrerInfo.referralCount || 0
+  }
+
+  const allResultsArray: [string, number, number, number | undefined][] = []
```
jophish, line 129: Why does this `undefined` need to be here?

scripts/referrerRevenueSummary.ts
```diff
@@ -0,0 +1,156 @@
...
+import { fetchReferralEvents, removeDuplicates } from './utils/referrals'
+
+async function getArgs() {
+  const argv = await yargs
+    .env('')
+    .option('protocol', {
+      alias: 'p',
+      description: 'protocol that the referrals are for',
+      demandOption: true,
+      choices: protocols,
```
silasbw, line 17: We don't actually know these, right?  Like, the way we do this is my manually submitting a tx (e.g., via a bloc explorer) vs editing a typescript file and running the script.  Remove `choices` and return an empty result if the protocol isn't registered?

MuckT: Setting the protocol is required because of how the `protocolFilter` and `calculateRevenueHandlers` are used. Is the hope that we would make this optional?

MuckT:
Closing as we will work to have this tracked on-chain.

PR #29, opened 2025-02-13
feat(registry): Add some quality-of-life changes to the contract

jophish:
This PR contains a handful of QOL changes to the contract. These include:

* Updating referrer/protocol identifiers to be `bytes32` instead of `string`. This makes parsing events indexed with these values easier.
* Adding a method to check whether a user is registered with a particular protocol, as well as a method for batching these lookups across multiple protocols.
* Adding a batch feature for registering referrals for users.

Notably, this does _not_ make `registerReferral` (or the new `registerReferrals`) idempotent, since the new `getUserRegistration` and `getUserRegistrations` obviate the need for idempotency, I think. The intended pattern for registering users would be as follows:
* Call `getUserRegistrations` on the user you'd like register, for all the protocols you'd like to register them for.
* For any protocols for which a `false` was returned, call `registerReferrals` passing them as an argument.
* `registerReferrals` should _only_ revert if the referrer themselves is not registered, or if the user is out of gas, but _not_ because the user is already registered to some protocol.

I think this is a fine solution for now; it's nice to have granular revert reasons, should we ever choose/figure out how to use them. There is a _small_ (extremely unlikely) race condition with this approach, where if some _other_ referrer registers a user between the `getUserRegistrations` call and `registerReferrals` call, `registerReferrals` could revert, though I'm not too worried about this, especially since the client would just repeat the steps above later and things would be OK.


contracts/Registry.sol
```diff
@@ -99,12 +99,35 @@ contract Registry is AccessControlDefaultAdminRules {
...
+      }
+      return registrations;
+  }
+
+  function getUserRegistration(
+                                 address userAddress,
+                                 bytes32 protocolId
+  ) external view returns (bool) {
+      return _userInfoByProtocol[protocolId][userAddress].timestamp != 0;
+  }
```
jeanregisser, line 118: Given the arguments and return value, what do you think of this?
Also I think 1 function that takes an array is enough.

```suggestion
  function isUserRegistered(
                                 address userAddress,
                                 bytes32[] calldata protocolIds
  ) external view returns (bool[] memory) {
      bool[] memory registeredStatuses = new bool[](protocolIds.length);
      for (uint256 i = 0; i < protocolIds.length; i++) {
          registeredStatuses[i] = _userInfoByProtocol[protocolId][userAddress].timestamp != 0;
      }
      return registeredStatuses;
  }
```

In addition, this is slightly more efficient because there's no call (though doesn't matter that much for a view call).

contracts/Registry.sol
```diff
@@ -99,12 +99,35 @@ contract Registry is AccessControlDefaultAdminRules {
...
+  function _registerReferral(
+                             bytes32 referrerId,
+                             bytes32 protocolId,
+                             address userAddress
+  ) internal {
+      _referrerIsRegistered = false;
+      for (uint256 i = 0; i < _referrerIdToProtocolIds[referrerId].length; i++) {
       if (
         keccak256(abi.encodePacked(_referrerIdToProtocolIds[referrerId][i])) ==
         keccak256(abi.encodePacked(protocolId))
```
jeanregisser, line 133: We can now do this directly which is more efficient:
```suggestion
        _referrerIdToProtocolIds[referrerId][i] ==
        protocolId
```

contracts/Registry.sol
```diff
@@ -99,12 +99,35 @@ contract Registry is AccessControlDefaultAdminRules {
...
+
+  // This registration functionality needs to be inside of an internal helper,
+  // rather than in the external `registerReferral` directly, since the batch
+  // `registerReferrals` must use this logic as a subroutine, and nested internal
+  // contract calls modify the value of `msg.sender`.
+  function _registerReferral(
+                             bytes32 referrerId,
+                             bytes32 protocolId,
+                             address userAddress
+  ) internal {
```
jeanregisser, line 128: Maybe we should only expose the batch version?

contracts/Registry.sol
```diff
@@ -99,12 +99,35 @@ contract Registry is AccessControlDefaultAdminRules {
...
+      _referrerIsRegistered = false;
```
jeanregisser, line 129: Why don't we use a local variable?

jeanregisser (commented):
Thanks! 👍

contracts/Registry.sol
jophish: Oh, weird -- this was in the original code, but I don't see a reason why it needs to be attached to the contract state. Updated!

contracts/Registry.sol
jophish: Oh :facepalm: missed this! Thanks for the catch

contracts/Registry.sol
jophish: Sure, I think it's fine to just expose batch methods for everything. It also means we can remove this indirection.

jeanregisser (approved):
Awesome 🚀

PR #32, opened 2025-02-19
chore: update abi and tests

kathaypacific:
This PR does a couple of small things:
- update the abi to account for the latest contract changes
- update the Registry tests to use `stringToHex` for the byte32 type conversion (`toBytes` returns a `ByteArray` which is causes typescript errors when used with `encodeFunctionData`)
- ensure command to verify contract is always printed for the deploy script as folks may want to verify the contracts when deployed locally

Note that we need to add `as const` to the end of the abi variable, but I did not do that here because it generates too many errors (the funding scripts all need to be updated to account for the contract changes, I spent some time on this but wasn't sure I was making the best changes)

### DIFF
--- a/abis/Registry.ts
+++ b/abis/Registry.ts
@@ -2,14 +2,78 @@ export const registryContractAbi = [
   {
     inputs: [
       {
-        internalType: 'string',
-        name: 'protocolId',
-        type: 'string',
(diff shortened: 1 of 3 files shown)


jeanregisser (approved):
🚀

PR #33, opened 2025-03-05
refactor(utils): make more generic functions for fetching blockchain events

jh2oman:
Creates a more generic `fetchEvents` function to fetch all events of a certain event name from a specific contract for a specified time range.

Preparation for in flight tickets that also query events from different contracts
* https://linear.app/valora/issue/ENG-175/tvl-calculation-for-somm
* https://linear.app/valora/issue/ENG-176/transaction-volume-for-aerodrome


finnian0826 (approved):
This is super helpful, thanks!

PR #34, opened 2025-03-05
feat(aerodrome): Compute swap transaction volume

finnian0826:
Computes swap transaction volume for [this liquidity pool](https://basescan.org/address/0xb2cc224c1c9feE385f8ad6a55b4d94E92359DC59). To add support for more pools, the pool address just needs to be added to the `SUPPORTED_LIQUIDITY_POOL_ADDRESSES` constant.

The expected return type of the `calculateRevenue` function is supposed to be a single value, so it did not make sense to return in terms of WETH and USDC, so I just have it convert everything to USD (was able to just copy pricing code from the beefy stuff).

**Manual Testing:**

Used the following script, which found this transaction and parsed the values correctly. (Price endpoint not working so it did not make it to the conversion step, but the intermediate token value was correct in the `SwapEvent` object.
```
import { calculateRevenue } from './scripts/calculateRevenue/protocols/aerodrome'

const address = '0xAe30f674684C78C3c33A0e3c07be5ebfce25aD99'
const startTimestamp = new Date('2025-03-04T20:00:00.000Z')
const endTimestamp = new Date('2025-03-04T21:00:00.000Z')

calculateRevenue({ address, startTimestamp, endTimestamp })
  .then((result) => {
    console.log(result)
  })
  .catch((error) => {
    console.error(error)
  })
```
Relevant transaction: [basescan](https://basescan.org/tx/0xb4fd5efa403397700763759cd106167bad0318a7e374f40f3c7978932c0326fd)


scripts/calculateRevenue/protocols/aerodrome/constants.ts
```diff
@@ -0,0 +1,9 @@
+import { Address } from 'viem'
+import { NetworkId } from '../../../types'
+
+export const SUPPORTED_LIQUIDITY_POOL_ADDRESSES: Address[] = [
+  '0xb2cc224c1c9feE385f8ad6a55b4d94E92359DC59',
+]
+export const AERODROME_NETWORK_ID = NetworkId['base-mainnet']
```
finnian0826, line 7: It seems like aerodrome only supports base, if there is a better way to have this information I can change it just lmk.

scripts/calculateRevenue/protocols/aerodrome/constants.ts
```diff
@@ -0,0 +1,9 @@
...
+
+export const TOKEN_AMOUNT_PRECISION = 8
```
finnian0826, line 9: Same value as what is done for beefy

scripts/calculateRevenue/protocols/aerodrome/getSwapEvents.ts
```diff
@@ -0,0 +1,59 @@
...
+  const allSwapEvents = await fetchEvents({
+    contract: swapContract,
+    networkId: AERODROME_NETWORK_ID,
+    eventName: 'Swap',
+    startTimestamp,
+    endTimestamp,
+  })
+  const filteredSwapEvents = allSwapEvents.filter(
+    (swapEvent) =>
+      (swapEvent.args as { recipient: string }).recipient === address,
```
jh2oman, line 27: i suggest [isAddressEqual](https://viem.sh/docs/utilities/isAddressEqual.html) to allow for the possibility of different casings

scripts/calculateRevenue/protocols/aerodrome/getSwapEvents.ts
```diff
@@ -0,0 +1,59 @@
...
+  )
+
+  const swapEvents: SwapEvent[] = []
+  const client = getViemPublicClient(AERODROME_NETWORK_ID)
+  const tokenId = await client.readContract({
```
jh2oman, line 32: nit: calling this variable tokenId is a bit misleading since tokenId has a proper definition within the rest of the valora ecosystem. `tokenAddress` ?

scripts/calculateRevenue/protocols/aerodrome/getSwapEvents.ts
```diff
@@ -0,0 +1,59 @@
...
+    const block = await client.getBlock({
+      blockNumber: swapEvent.blockNumber,
+    })
+    swapEvents.push({
+      timestamp: new Date(Number(block.timestamp * 1000n)),
+      amountInToken:
+        Number(
+          (((swapEvent.args as { amount0: bigint }).amount0 > 0n
+            ? (swapEvent.args as { amount0: bigint }).amount0
+            : -(swapEvent.args as { amount0: bigint }).amount0) *
```
jh2oman, line 50: this is just getting the absolute value of amount0, right? I think it might be cleaner to have this in a temporary variable like `const absAmount0`

scripts/calculateRevenue/protocols/aerodrome/getSwapEvents.ts
```diff
@@ -0,0 +1,59 @@
...
+            10n ** BigInt(TOKEN_AMOUNT_PRECISION)) /
+            10n ** tokenDecimals,
+        ) /
+        10 ** TOKEN_AMOUNT_PRECISION,
```
jh2oman, line 54: I think when @jophish does the PRECISION cutoff for the beefy code it is specifically at the conversion to USD step because for USD we don't necessarily care about fractions of fractions of a cent. 

I think doing it at this step is potentially impactful in a harmful way because you are cutting off information before it is converted to USD. For example with ETH we *do care* about very low numbers.

@jophish feel free to weigh in.

scripts/calculateRevenue/protocols/aerodrome/getSwapEvents.ts
finnian0826: Updated to be used in [f7e3682](https://github.com/divvi-xyz/divvi-protocol-v0/pull/34/commits/f7e3682fd7e5abe7f163ca12415d690d5c5648a8), however had to cast the `address` as type `Address`, since that is what is needed for the function and it is passed in as type `string` to `calculateRevenue`

scripts/calculateRevenue/protocols/aerodrome/getSwapEvents.ts
finnian0826: Updated in [f7e3682](https://github.com/divvi-xyz/divvi-protocol-v0/pull/34/commits/f7e3682fd7e5abe7f163ca12415d690d5c5648a8)

scripts/calculateRevenue/protocols/aerodrome/getSwapEvents.ts
finnian0826: pulled out into its own variable in [f7e3682](https://github.com/divvi-xyz/divvi-protocol-v0/pull/34/commits/f7e3682fd7e5abe7f163ca12415d690d5c5648a8) and cleaned up a bit

scripts/calculateRevenue/protocols/aerodrome/getSwapEvents.ts
finnian0826: Updated in [f7e3682](https://github.com/divvi-xyz/divvi-protocol-v0/pull/34/commits/f7e3682fd7e5abe7f163ca12415d690d5c5648a8) to more closely match what is done in the Beefy calculation

scripts/calculateRevenue/protocols/aerodrome/types.ts
```diff
@@ -0,0 +1,5 @@
+export type SwapEvent = {
+  timestamp: Date
+  amountInToken: number
```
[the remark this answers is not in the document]
finnian0826: Updated in [f7e3682](https://github.com/divvi-xyz/divvi-protocol-v0/pull/34/commits/f7e3682fd7e5abe7f163ca12415d690d5c5648a8)

jh2oman (approved):
Looks great!

PR #35, opened 2025-03-05
chore: add berachain

silasbw:


hardhat.config.ts
```diff
@@ -52,6 +52,11 @@ const config: HardhatUserConfig = {
       accounts,
       chainId: 8453,
     },
+    {
```
jophish, line 55: Can we also add an entry under `customChains` for etherscan/contract verification?

jophish (approved):

PR #47, opened 2025-03-13
feat: gas fees for blockchain networks

MuckT:
### Description

Gets all gas fees for registered users on a specific network - WIP

### Testing instructions

1. Grab an API key from https://envio.dev/app/api-tokens
2. Set your .env following `.env.example`.
3. Run the following script: 
```bash
yarn ts-node ./scripts/calculateNetworkRevenue.ts --protocolId beefy --networkId arbitrum-one

# Total gas used (Wei): 1104758315755000
# ✨  Done in 3.02s.
```


README.md
```diff
@@ -53,6 +53,15 @@ Calculates revenue for a list of users on a specific protocol between startTimes
...
 
+### Calculate Network Revenue
+
+Calculates the revenue for a network on a specific protocol
+
+```bash
+# startBlock & endBlock are optional
+yarn ts-node ./scripts/calculateNetworkRevenue.ts --protocolId Beefy --networkId arbitrum-one
+```
+
```
jh2oman, line 64: Why is this a separate script? Shouldn't this be a part of calculateRevenue where 'optimism' or 'celo' are protocol options?

scripts/calculateNetworkRevenue.ts
```diff
@@ -0,0 +1,168 @@
...
+        sum + BigInt(tx.gasUsed ?? 0) * BigInt(tx.gasPrice ?? 0),
+      0n,
+    )
+  } catch (error) {
+    console.log('Error fetching gas usage: ', error)
+    return 0n
+  }
+}
+
+async function getUsersForProtocol({
```
jh2oman, line 81: can you extend the existing fetchReferrals script instead of writing a new one?

README.md
jh2oman: > AC:

> Integrate code with https://github.com/divvi-xyz/divvi-protocol-v0/blob/main/scripts/calculateRevenue.ts

scripts/calculateNetworkRevenue.ts
```diff
@@ -0,0 +1,168 @@
...
+    ...(endBlock !== null && { toBlock: endBlock }),
+  }
+
+  try {
+    const response: QueryResponse = await client.get(query)
+    return response.data.transactions.reduce(
+      (sum: bigint, tx) =>
+        sum + BigInt(tx.gasUsed ?? 0) * BigInt(tx.gasPrice ?? 0),
+      0n,
+    )
```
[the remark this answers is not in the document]
MuckT: I tried the stream method and it resulted in significantly slower execution time: 30s vs. the 3s previously. Instead, we can track the archiveHeight vs. nextBlock to check that we've hit the most recent block: 4b69b4eee53e7cb602e670ace46aa0809e6a8a98.

README.md
MuckT: Based on what the AC states below it sounds like we want to provide a network, protocol ID, and optional block range. 

> Given a blockchain network (e.g., Celo, Polygon, Optimism, etc.) a frontend ID/referrer ID, and block range we're able to calculate the sum of network revenue attributable to the frontend in that block range

scripts/calculateNetworkRevenue.ts
MuckT: I was able to use the functions from `fetchReferralEvents` and `removeDuplicates` instead of creating my own to avoid duplication.

MuckT:
Closing in favor of #56.

PR #51, opened 2025-03-13
fix(scripts): clean up fetchReferrals and calculateRevenue scripts

jh2oman:
Several fixes and improvements to the scripts so that now you can actually run the following and get a result
```bash
$ yarn ts-node ./scripts/fetchReferrals.ts --protocol beefy
$ yarn ts-node ./scripts/calculateRevenue.ts --protocol beefy --startTimestamp 1740013389000 --endTimestamp 1741899467000
```


scripts/calculateRevenue.ts
```diff
@@ -6,55 +6,60 @@ import yargs from 'yargs'
 import { protocols, Protocol } from './types'
 
 async function main(args: ReturnType<typeof parseArgs>) {
-  const eligibleAddresses = parse(
-    readFileSync(args['input-addresses'], 'utf-8').toString(),
-    { skip_empty_lines: true },
-  ).flat()
-  const handler = calculateRevenueHandlers[args['protocol-id'] as Protocol]
+  const inputFile = args['input-file'] ?? `${args['protocol']}-referrals.csv`
+  const outputFile = args['output-file'] ?? `${args['protocol']}-revenue.csv`
```
satish-ravi, line 10: nit: yargs should give you these in camel case, so you can do `args.inputFile`, `args.outputFile`, etc which I find a little easier to read.

scripts/calculateRevenue.ts
```diff
@@ -63,14 +68,14 @@ function parseArgs() {
     .option('start-timestamp', {
       alias: 's',
       description:
-        'timestamp at which to start checking for revenue (since epoch)',
+        'timestamp at which to start checking for revenue (milliseconds since epoch)',
```
satish-ravi, line 71: maybe easier to do iso ts? or actually anything Date(..) can accept? E.g., if we don't care about times, can we just pass the date in `yyyy-mm-dd`?

scripts/calculateRevenue/protocols/beefy/index.ts
```diff
@@ -113,7 +113,7 @@ export async function calculateVaultRevenue(
       BigInt(vaultTvl * 10 ** 18)
     const partialUsdContribution =
       Number(
-        (BigInt(tokenPriceUsd * 10 ** REVENUE_USD_PRECISION) *
+        (BigInt((tokenPriceUsd * 10 ** REVENUE_USD_PRECISION).toFixed(0)) *
```
satish-ravi, line 116: could also use parseUnits / formatUnits from viem for these conversions but those may need string handling

satish-ravi (approved):

scripts/calculateRevenue/protocols/beefy/index.ts
jh2oman: thats a better solution. thanks!

PR #54, opened 2025-03-17
chore: Update supported aerodrome pools to include top 8 pools by TVL/volume

finnian0826:
For [this issue](https://linear.app/valora/issue/ENG-203/support-more-pools-on-aerodrome)

Top 5 by [TVL](https://aerodrome.finance/liquidity?sort=tvl&asc=false)
Topy 5 by [volume](https://aerodrome.finance/liquidity?sort=volume&asc=false)


jh2oman:
@finnian0826 I say add superOETHb. I don't think we should be very heavy handed with our judgment of what to include or not. Its their (Aerodrome's) product

finnian0826:
> @finnian0826 I say add superOETHb. I don't think we should be very heavy handed with our judgment of what to include or not. Its their (Aerodrome's) product

I didn't initially add it because then it would show up in Valora, but I see now I can set `isSwappable` to `false` and then that should be okay. I'll add it to address-metadata and then add the pool here

scripts/calculateRevenue/protocols/aerodrome/constants.ts
```diff
@@ -3,6 +3,13 @@ import { NetworkId } from '../../../types'
 
 export const SUPPORTED_LIQUIDITY_POOL_ADDRESSES: Address[] = [
   '0xb2cc224c1c9feE385f8ad6a55b4d94E92359DC59',
+  '0x6cDcb1C4A4D1C3C6d054b27AC5B77e89eAFb971d',
```
jophish, line 6: Can you add a comment for each of these just mentioning what they are?

finnian0826: added in [57549e2](https://github.com/divvi-xyz/divvi-protocol-v0/pull/54/commits/57549e2cf6f82d451e554ab09ee12418ef5a70ed)!

jh2oman (approved):
Update the PR title to reflect the number of pools that are being added

PR #56, opened 2025-03-18
feat: calculate gas fees for blockchain networks

MuckT:
### Description

Gets all transaction fees for registered users on a specific network.

### Testing instructions

1. Grab an API key from https://envio.dev/app/api-tokens
2. Set your .env following `.env.example`.
3. Run the following scripts: 
```bash
# Generates arbitrum-referrals.csv - may want to use `--protocol beefy` and rename the resulting file to get data.
yarn ts-node ./scripts/fetchReferrals.ts --protocol arbitrum

# Run calculateRevenue.ts with timestamps from March 18th 2022 to Marth 13th 2025
yarn ts-node ./scripts/calculateRevenue.ts --protocol arbitrum --startTimestamp 1647595987000 --endTimestamp 1741899467000
```

Resulting `arbitrum-revenue.csv`
```
investor,0x15B5f5FE55704140ce5057d85c28f8b237c1Bc53,1740013389
```

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,3 +4,4 @@ DEFENDER_API_KEY=<OPENZEPPELIN_DEFENDER_API_KEY>
 DEFENDER_API_SECRET=<OPENSEPPELIN_DEFENDER_API_SECRET>
 SECRET_NAMES=projects/1027349420744/secrets/funding-layer/versions/1
 ALCHEMY_KEY=<ALCHEMY_KEY>
+HYPERSYNC_API_KEY=<https://docs.envio.dev/docs/HyperSync/api-tokens#generating-api-tokens>
--- a/scripts/calculateRevenue/protocols/celo/index.ts
+++ b/scripts/calculateRevenue/protocols/celo/index.ts
@@ -1,6 +1,6 @@
 import { NetworkId } from '../../../types'
 import { getNearestBlock } from '../utils/events'
-import { fetchTotalGasUsed } from '../utils/networks'
+import { fetchTotalTransactionFees } from '../utils/networks'
 
 export async function calculateRevenue({
   address,
@@ -20,7 +20,7 @@ export async function calculateRevenue({
     endTimestamp,
   )
 
-  return await fetchTotalGasUsed({
+  return await fetchTotalTransactionFees({
     networkId: NetworkId['celo-mainnet'],
     users: [address],
     startBlock,
--- a/scripts/calculateRevenue/protocols/celo/index.ts
+++ b/scripts/calculateRevenue/protocols/celo/index.ts
@@ -1,0 +1,29 @@
+import { NetworkId } from '../../../types'
+import { getNearestBlock } from '../utils/events'
+import { fetchTotalGasUsed } from '../utils/networks'
+
+export async function calculateRevenue({
+  address,
+  startTimestamp,
+  endTimestamp,
+}: {
+  address: string
+  startTimestamp: Date
+  endTimestamp: Date
+}): Promise<number> {
+  const startBlock = await getNearestBlock(
+    NetworkId['celo-mainnet'],
+    startTimestamp,
+  )
+  const endBlock = await getNearestBlock(
+    NetworkId['celo-mainnet'],
+    endTimestamp,
+  )
+
+  return await fetchTotalGasUsed({
+    networkId: NetworkId['celo-mainnet'],
+    users: [address],
+    startBlock,
+    endBlock,
+  })
+}
--- a/scripts/calculateRevenue/protocols/utils/networks.ts
+++ b/scripts/calculateRevenue/protocols/utils/networks.ts
@@ -39,6 +39,11 @@ export async function fetchTotalGasUsed({
       for (const tx of response.data.transactions) {
         totalGasUsed += Number(tx.gasUsed ?? 0) * Number(tx.gasPrice ?? 0)
       }
+
+      // Check if we've reached the desired end block to avoid an unnecessary request
+      if (endBlock && query.fromBlock >= endBlock) {
+        hasMoreBlocks = false
+      }
     }
   } catch (error) {
     console.error('Error fetching transactions:', error)
--- a/scripts/calculateRevenue/protocols/utils/networks.ts
+++ b/scripts/calculateRevenue/protocols/utils/networks.ts
@@ -13,7 +13,6 @@ export async function fetchTotalGasUsed({
   startBlock?: number
   endBlock?: number
 }): Promise<number> {
-  let fromBlock = startBlock ?? 0
   let totalGasUsed = 0
   let hasMoreBlocks = true
 
@@ -25,31 +24,22 @@ export async function fetchTotalGasUsed({
       fieldSelection: {
         transaction: [TransactionField.GasUsed, TransactionField.GasPrice],
       },
-      fromBlock,
+      fromBlock: startBlock ?? 0,
       ...(endBlock && { toBlock: endBlock }),
     }
 
-    do {
+    while (hasMoreBlocks) {
       const response: QueryResponse = await client.get(query)
-
-      if (
-        response.nextBlock <= fromBlock ||
-        !response.data.transactions.length
-      ) {
+      if (response.nextBlock === query.fromBlock) {
         hasMoreBlocks = false
+      } else {
+        query.fromBlock = response.nextBlock
       }
 
       for (const tx of response.data.transactions) {
         totalGasUsed += Number(tx.gasUsed ?? 0) * Number(tx.gasPrice ?? 0)
       }
-
-      fromBlock = response.nextBlock
-      query.fromBlock = fromBlock
-
-      if (endBlock && fromBlock >= endBlock) {
-        hasMoreBlocks = false
-      }
-    } while (hasMoreBlocks)
+    }
   } catch (error) {
     console.error('Error fetching transactions:', error)
     return 0
--- a/scripts/calculateRevenue/protocols/utils/networks.ts
+++ b/scripts/calculateRevenue/protocols/utils/networks.ts
@@ -1,0 +1,59 @@
+import { QueryResponse, TransactionField } from '@envio-dev/hypersync-client'
+import { getHyperSyncClient } from '../../../utils'
+import { NetworkId } from '../../../types'
+
+export async function fetchTotalGasUsed({
+  networkId,
+  users,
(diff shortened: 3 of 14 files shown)


scripts/calculateRevenue/protocols/celo/index.ts
```diff
@@ -0,0 +1,29 @@
...
+    startTimestamp,
+  )
+  const endBlock = await getNearestBlock(
+    NetworkId['celo-mainnet'],
+    endTimestamp,
+  )
+
+  return await fetchTotalGasUsed({
+    networkId: NetworkId['celo-mainnet'],
+    users: [address],
```
MuckT, line 25: We could adjust `fetchTotalGasUsed` to take just one user, but as this function can support multiple users, this retains functionality while conforming to the expected `calculateRevenue` parameters.

.env.example
```diff
@@ -4,3 +4,4 @@ DEFENDER_API_KEY=<OPENZEPPELIN_DEFENDER_API_KEY>
 DEFENDER_API_SECRET=<OPENSEPPELIN_DEFENDER_API_SECRET>
 SECRET_NAMES=projects/1027349420744/secrets/funding-layer/versions/1
 ALCHEMY_KEY=<ALCHEMY_KEY>
+HYPERSYNC_API_KEY=<https://docs.envio.dev/docs/HyperSync/api-tokens#generating-api-tokens>
```
jophish, line 7: Can we add this to the GCP `funding-layer` secret?

scripts/calculateRevenue/protocols/utils/networks.ts
```diff
@@ -0,0 +1,59 @@
...
+    const query = {
+      transactions: [{ from: users }],
+      fieldSelection: {
+        transaction: [TransactionField.GasUsed, TransactionField.GasPrice],
+      },
+      fromBlock,
+      ...(endBlock && { toBlock: endBlock }),
+    }
+
+    do {
```
jophish, line 32: Can we just use `while` here? We never actually take advantage of the at-least-once execution semantics of `do while` since `hasMoreBlocks` will always be true initially.

scripts/calculateRevenue/protocols/utils/networks.ts
```diff
@@ -0,0 +1,59 @@
...
+
+export async function fetchTotalGasUsed({
+  networkId,
+  users,
+  startBlock,
+  endBlock,
+}: {
+  networkId: NetworkId
+  users: string[]
+  startBlock?: number
```
jophish, line 13: Do we ever expect either of `startBlock` or `endBlock` to be `undefined`? It seems like accidentally setting these to undefined could make the query balloon in size by a bunch.

scripts/calculateRevenue/protocols/utils/networks.ts
```diff
@@ -0,0 +1,59 @@
...
+  endBlock?: number
+}): Promise<number> {
+  let fromBlock = startBlock ?? 0
+  let totalGasUsed = 0
+  let hasMoreBlocks = true
+
+  try {
+    const client = getHyperSyncClient(networkId)
+
+    const query = {
```
jophish, line 23: Is this query fetching all TXs a user has made on the network across a given block range?

scripts/calculateRevenue/protocols/utils/networks.ts
```diff
@@ -0,0 +1,59 @@
...
+      fromBlock,
+      ...(endBlock && { toBlock: endBlock }),
+    }
+
+    do {
+      const response: QueryResponse = await client.get(query)
+
+      if (
+        response.nextBlock <= fromBlock ||
+        !response.data.transactions.length
```
jophish, line 37: How does pagination work for this endpoint? Is the page size equal to a fixed number of blocks, or a fixed number of TXs? In other words, is it possible for the endpoint to return 0 transactions in a response? Like if we requested block range 0-1000, and there were no TXs between 0-100 (but transactions afterwards), would the query return 0 TXs the first time? If so, the `!response.data.transactions.length` condition would cause `hasMoreBlocks` to be set prematurely.

scripts/calculateRevenue/protocols/utils/networks.ts
```diff
@@ -0,0 +1,59 @@
...
+      ) {
+        hasMoreBlocks = false
+      }
+
+      for (const tx of response.data.transactions) {
+        totalGasUsed += Number(tx.gasUsed ?? 0) * Number(tx.gasPrice ?? 0)
+      }
+
+      fromBlock = response.nextBlock
```
jophish, line 46: nit: I think you can save on this assignment here by just replacing all `fromBlock` usages within the `while` loop with `query.fromBlock`.

scripts/calculateRevenue/protocols/utils/networks.ts
```diff
@@ -0,0 +1,59 @@
...
+      query.fromBlock = fromBlock
+
+      if (endBlock && fromBlock >= endBlock) {
```
jophish, line 49: It seems like the intent of this check is to determine if we've reached "the end" of the results. In the case of `endBlock` not being set, this will never be true. Is `response.nextBlock` undefined when we've caught up with the current state of the chain? If so, can we change this to e.g. `(endBlock && fromBlock >= endBlock) || (!endBlock && !response.nextBlock))`?

jophish (commented):
Can we also add some tests for `networks.ts`?

scripts/calculateRevenue/protocols/utils/networks.ts
MuckT: Hypersync can handle it: [Get All ERC-20 Transfers for an Addres](https://docs.envio.dev/docs/HyperSync/hypersync-curl-examples#get-all-erc-20-transfers-for-an-address).

scripts/calculateRevenue/protocols/utils/networks.ts
MuckT: Good catch! Pagination constraints can be set either via `maxNumBlocks` or `maxNumTransactions`; however, the default numbers for these values [were not found](https://github.com/enviodev/hypersync-client-node/blob/ef55e364d26845c612a5233fdf8a2c182b2ffefa/index.d.ts#L242-L312). In testing are there are cases where transactions return an empty array with blocks were remaining to be processed. I'll remove this check.

scripts/calculateRevenue/protocols/utils/networks.ts
```diff
@@ -0,0 +1,49 @@
...
+    while (hasMoreBlocks) {
+      const response: QueryResponse = await client.get(query)
+      if (response.nextBlock === query.fromBlock) {
+        hasMoreBlocks = false
+      } else {
+        query.fromBlock = response.nextBlock
+      }
+
+      for (const tx of response.data.transactions) {
+        totalGasUsed += Number(tx.gasUsed ?? 0) * Number(tx.gasPrice ?? 0)
```
silasbw, line 40: naming nit: `totalGasUsed` looks like it's actually the sum of transaction fees.  How about naming this `totalTransactionFees`?

silasbw, line 40: Should these be BitInts?  I imagine that's why they come back as strings instead of numbers.

scripts/calculateRevenue/protocols/utils/networks.ts
```diff
@@ -0,0 +1,49 @@
...
+      }
+    }
+  } catch (error) {
+    console.error('Error fetching transactions:', error)
```
silasbw, line 44: Why not throw?

scripts/calculateRevenue/protocols/utils/networks.ts
```diff
@@ -0,0 +1,49 @@
+import { QueryResponse, TransactionField } from '@envio-dev/hypersync-client'
+import { getHyperSyncClient } from '../../../utils'
+import { NetworkId } from '../../../types'
+
+export async function fetchTotalGasUsed({
```
silasbw, line 5: same naming nit here: this returns total transaction fees, right?  not gas

scripts/calculateRevenue/protocols/utils/networks.ts
MuckT: [CalculateRevenueFn](http://github.com/divvi-xyz/divvi-protocol-v0/blob/5dbfc211271f64cb670199b6451b4f9eafb6c14a/scripts/types.ts#L27-L31) has the return type of `Promise<number>`, but maybe we should change this return type for all `CalculateRevenueFn` functions separate PR?

scripts/calculateRevenue/protocols/utils/networks.ts
MuckT: Changed in bae945861c4e13f39ca8496a940ead98c0ea0a4a!

scripts/calculateRevenue/protocols/utils/networks.ts
MuckT: Changed in e36298d18e323ff1455de7bbab21b05794ed1029!

scripts/calculateRevenue/protocols/utils/networks.ts
MuckT: Changed in e36298d18e323ff1455de7bbab21b05794ed1029.

scripts/calculateRevenue/protocols/utils/networks.ts
MuckT: Changed in 1676e203764a05adeb799a89ebf8d5399dd40d3b.

scripts/calculateRevenue/protocols/utils/networks.ts
MuckT: I fixed this in 0e858e96c6172a661ef39b7f3595cac62e8615b3 when `endBlock` is provided. If `endBlock` is missing, the function makes one additional request before exiting the loop. The missing `endBlock` shouldn’t happen since the CLI argument `endTimestamp` is required, but I've added a test for this regardless.

.env.example
MuckT: Added!

scripts/calculateRevenue/protocols/utils/networks.ts
silasbw: OK, good suggestion.  Leave it as Number for now :+1:

silasbw (approved):

jophish (approved):

PR #59, opened 2025-03-18
feat: reward pool contract

bakoushin:
### Description

Upgradeable reward pool contract.

High-level overview:
* supports a single native or an ERC-20 token for distributing rewards
* stores reward amount of each user on-chain
* requires the manager to periodcally upload this data in batches
* allows claiming rewards as long as the funds are available

Allows a _manager_ to:
* `deposit` funds
*  `withdraw` funds after a certain timelock defined at contract deployment time
* `extendTimelock` in the future if needed
* `addRewards` in batches, listing user addresses and the reward amounts for each user
   * each batch also includes `rewardFunctionArgs`, referring to the arguments used for calculating rewards
 * `rescueToken` to extract any extra tokens accidentally landed in the contract

Allows users to:
* see `pendingRewards` for their address
* `claimReward` for their address

Allows anyone to:
* see `rewardFunctionId`, referring to a specific commit on GitHub containing the specific reward function logic (set at deployment)
* see `poolBalance`
* see  `totalPendingRewards`
* see `poolToken` address
* see `timelock` timestamp

> [!IMPORTANT] 
> The contract intentionally doesn't have any protection against adding more reward entitlements than the available funds. Users can claim only as long as funds are available.

### Expected usage

1. Contract is deployed by the engineering team, specifying:
    a) pool token
    b) reward function commit
    c) manager address
    d) initial timelock
2. The manager deposits funds into the contract
3. Manager periodically adds reward entitlements for users
4. Users can claim the rewards they are entitled to as long as there are funds available
5. After the timelock expires, the manager withdraws the rest of the funds
6. If needed, the manager could:
    a) extend the timelock
    b) deposit more funds
7. The manager could withdraw any other non-pool tokens if they land in this contract for some reason (e.g. accidental transfer).

### Test plan

* Add unit test in a separate PR

### Related issues

* Related to ENG-204

### Possible future considerations

* A _factory contract_ to allow managers to deploy specific pools themselves
* Consider _reward epochs_ to store only the epoch's merkle root on-chain instead of tracking each user balance


contracts/RewardPool.sol
```diff
@@ -0,0 +1,254 @@
...
+            if (amounts[i] == 0) revert AmountMustBeGreaterThanZero();
+            
+            pendingRewards[users[i]] += amounts[i];
+            totalPendingRewards += amounts[i];
+
+            emit AddReward(users[i], amounts[i], rewardsAllocationId);
+
+            unchecked {
+                i++;
+            }
```
jh2oman, line 170: Do we really want this to be unchecked? I feel like if i is so large that it overflows we will want to revert the call, not wrap.

contracts/RewardPool.sol
```diff
@@ -0,0 +1,254 @@
...
+        
+        _transferPoolToken(msg.sender, amount);
+        emit Withdraw(amount);
+    }
+    
+    /**
+     * @dev Increases amounts available for users to claim
+     * @param users Array of user addresses
+     * @param amounts Array of amounts to allocate for each user
+     * @param rewardsAllocationId Bytes32 identifier of the rewards allocation (e.g. git commit hash)
```
jh2oman, line 147: I wonder if there is an easy way to make this more transparent. Perhaps a bytes32[] array of parameters used to call the reward function? That way anybody should be able to verify the results of the rewards function call.

contracts/RewardPool.sol
bakoushin: yes, that's right from the theoretical standpoint.

the assumption here is that it'll unlikely hit `i < 2^256` in this loop so that we can save a little gas. even if someone will attempt to supply an array of that size, they should hit the gas limit rather than the overflow.

although I agree that given this is an experimental contract, we should keep its code simpler. will revert to a classic loop.

contracts/RewardPool.sol
bakoushin: yes, that makes sense. what could those parameters look like? from a naive standpoint, they are just the block numbers framing the calculation interval. or could they be more complex than that?

jh2oman: currently the calculation is just a time interval framing the calculation like you suggest, in milliseconds since epoch though.

contracts/RewardPool.sol
```diff
@@ -0,0 +1,254 @@
...
+    using SafeERC20 for IERC20;
+    
+    // Constants
+    address public constant NATIVE_TOKEN_ADDRESS = 0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE;
+    bytes32 public constant MANAGER_ROLE = keccak256("MANAGER_ROLE");
+    
+    // Init variables
+    address public poolToken;
+    bool public isNativeToken;
+    bytes32 public rewardFunctionId;
```
jeanregisser, line 25: Should these be `immutable`?
Or do we plan to allow changing them?

contracts/RewardPool.sol
jeanregisser: FWIW, the loop guard ensures this unchecked increment can't overflow, since `i < usersLength`. And `userLength` is `uint256` too.

But the readability point still holds.

contracts/RewardPool.sol
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    /**
+     * @dev Internal function to set the timelock
+     * @param blockNumber Block number when withdrawals will be allowed
+     */
+    function _setTimelock(uint256 blockNumber) internal {
+        if (blockNumber <= block.number) revert TimelockMustBeInTheFuture();
+        timelock = blockNumber;
+    }
```
jeanregisser, line 202: Just checking, do we really want a block number based lock?
We could use `block.timestamp` instead and have time-based lock.

contracts/RewardPool.sol
```diff
@@ -0,0 +1,254 @@
...
+
+    /**
+     * @dev Extends the timelock for manager withdrawals
+     * @param blockNumber Future block number when withdrawals will be allowed
+     * @notice Allowed only for address with MANAGER_ROLE
+     */
+    function extendTimelock(uint256 blockNumber) external onlyRole(MANAGER_ROLE) {
+        _setTimelock(blockNumber);   
+        emit TimelockExtended(blockNumber);
+    }
```
jeanregisser, line 111: `_setTimelock` only checks `blockNumber <= block.number` as an invalid scenario. 
This means that `extendTimelock` could actually shorten the timelock if someone passes a value still larger than `block.number` but lower than the current timelock.

Maybe it should be called `updateTimelock`?

jeanregisser (commented):
Great work!! 🙌

contracts/RewardPool.sol
bakoushin: well, it is not recommended for the upgradeable contracts: https://docs.openzeppelin.com/upgrades-plugins/faq#why-cant-i-use-immutable-variables

also, who knows, we may want to upgrade some of them

contracts/RewardPool.sol
bakoushin: oh, that's a great catch, thank you!

i guess reducing the existing timelock shouldn't be allowed, will fix that

contracts/RewardPool.sol
bakoushin: yes, that makes sense to simplify how we reason about then it expires. especially given different chains have different block time. will refactor to timestamp

contracts/RewardPool.sol
bakoushin: since they are both numbers, maybe we could restrict the array to `uint256` so the numbers are formatted nicely in the block explorer?

contracts/RewardPool.sol
```diff
@@ -0,0 +1,286 @@
...
+
+  // Events
+  event PoolInitialized(
+    address indexed poolToken,
+    bytes32 rewardFunctionId,
+    uint256 timelock
+  );
+  event Deposit(uint256 amount);
+  event Withdraw(uint256 amount);
+  event TimelockExtended(uint256 newTimelock);
```
finnian0826, line 46: Would we want to include the current timelock in this event to be able to compare (not sure what this event is going to be used for down the line, but could be useful info).

contracts/RewardPool.sol
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  /**
+   * @dev Allows the manager to deposit funds for rewards
+   * @param amount Amount to deposit (required for ERC-20, informational for native token)
+   * @notice Allowed only for address with MANAGER_ROLE
+   */
+  function deposit(uint256 amount) external payable onlyRole(MANAGER_ROLE) {
+    if (isNativeToken) {
+      if (msg.value != amount) revert AmountMismatch(amount, msg.value);
```
finnian0826, line 145: Just for my understanding, is there an implicit transfer done when `deposit` is called with the native token? Something similar to the `safeTransferFrom` done for non native tokens below?

jeanregisser (approved):
🚀

contracts/RewardPool.sol
bakoushin: yes, the native value is transferred via the `value` prop of the transaction:

```typescript
const txHash = await walletClient.writeContract({
  address: rewardPoolAddress,
  abi: rewardPoolAbi,
  functionName: 'deposit',
  args: [depositAmount], 
  value: depositAmount // <-- this will transfer the native value
});
```

contracts/RewardPool.sol
bakoushin: great idea! thanks, will add this

PR #62, opened 2025-03-19
feat(somm): Filter referrals for Somm

finnian0826:
Check that the users TVL across all vaults is 0 at the time of the referral event, and that they have interacted with Somm at least once since the referral (this is done for beefy, so also did it here).

Memoization added to `getEvents` function.


scripts/protocolFilters/somm.ts
```diff
@@ -1,6 +1,44 @@
...
+      abi: erc20Abi,
+      client,
+    })
+    const currentLPTokenBalance = await vaultContract.read.balanceOf([address])
+    const tokenDecimals = await vaultContract.read.decimals()
+    const currentTvl = Number(formatUnits(currentLPTokenBalance, tokenDecimals))
+    const tvlEvents = await getEvents({
+      address,
+      vaultInfo,
+      startTimestamp: eventDate,
```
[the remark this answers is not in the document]
finnian0826: There shouldn't be multiple referral events for the same address, so I agree I'm not sure what benefit the memoization adds. However the Eng notes of the [ticket](https://linear.app/valora/issue/ENG-200/somm-add-filter-function-for-valid-referrals) say to memoize similar to how it is done in the beefy implementation, and I wasn't sure what to memoize other than the event fetching?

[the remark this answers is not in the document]
jophish: Hmm, with the way things are currently structured, yeah, I don't see how the memoization is getting us anything.

In theory, the underlying [getEvents](https://github.com/divvi-xyz/divvi-protocol-v0/blob/main/scripts/calculateRevenue/protocols/somm/getEvents.ts#L39) function calls [fetchEvents](https://github.com/divvi-xyz/divvi-protocol-v0/blob/main/scripts/calculateRevenue/protocols/utils/events.ts#L63), which _is_ memoized. From call-to-call, for a single vault, the only parameters to `fetchEvents` that will vary are `startTimestamp` and `endTimestamp`; `startTimestamp` will _definitely_ vary each time, since this depends on the timestamp in the `ReferralEvent` object. 

This means that `fetchEvents` will be fetching big overlaps of the same vault events for all addresses many times. Ideally, we actually call `fetchEvents` once per vault, and just filter those memoized results by address in `getEvents`.

One way we can do this is by passing fixed start/end timestamps to `getEvents` here. For example, we could do something like:
* For each vault, find out the timestamp/block when it was deployed (so we know the earliest possible time anyone could have interacted with it)
* In this function, get events for each vault/address combo from that deploy time up to the referral event timestamp.
* Compare TVL at referral to current TVL; if TVL at referral is 0 and current TVL > 0, then it's a valid referral.

This will have the effect of actually meaningfully memoizing the event fetching, I'm pretty sure. Just one idea!

scripts/protocolFilters/somm.ts
```diff
@@ -1,6 +1,44 @@
...
+      endTimestamp: nowDate,
+    })
+    numTvlEvents += tvlEvents.length
+    const tvlAtReferral =
+      currentTvl - tvlEvents.reduce((acc, event) => acc + event.amount, 0)
+    if (tvlAtReferral > 0) {
```
jophish, line 39: Do we care that their TVL at referral was 0? In theory, this means that a user could have interacted with Somm a bunch before the referral event, as long as their TVL was exactly 0 when the referral happened. Do we instead want to check that their # of transactions with the vault was 0 before the referral event?

finnian0826: In talking with @jh2oman it sounds like we only want to check if the user has TVL at the time of referral. I think the idea is that even if they had interacted with Somm before, but then were "re-activated" by some referrer then that referral is still valid, but Jacob can correct me if I'm wrong.

jh2oman: Number of transactions with the vault would definitely be a more correct response. I had initially written it off as probably too intensive and settled with TVL 0 as an alternative. With hypersync though stepping through all events for the history of the vault seems more possible.

jh2oman: I'm pretty sure for this stage either approach is acceptable, we might as well go with the one we have already implemented in the interest of speed.

finnian0826: That makes sense, so keep this as is for now to have something ready to go for v0, and then have a follow-up ticket during v1 to do a "better" check (might get some for free from [this ticket](https://linear.app/valora/issue/ENG-194/filter-for-valid-aerodrome-referrals) that will use hypersync)

scripts/protocolFilters/somm.ts
```diff
@@ -1,6 +1,44 @@
...
+  for (const vaultInfo of vaultsInfo) {
+    const client = getViemPublicClient(vaultInfo.networkId)
+    const vaultContract = getContract({
+      address: vaultInfo.vaultAddress,
+      abi: erc20Abi,
+      client,
+    })
+    const currentLPTokenBalance = await vaultContract.read.balanceOf([address])
+    const tokenDecimals = await vaultContract.read.decimals()
+    const currentTvl = Number(formatUnits(currentLPTokenBalance, tokenDecimals))
```
jophish, line 29: Is there any reason to cast the bigint to a number here? can we just deal with bigints all the way through here?

finnian0826: The events queried for a few lines below this returns a list of `TvlEvent`, which has an `amount` that is type `number`. So the `tvlEvents.reduce` will be of type `number` and we subtract that from `currentTvl` so I thought that should be cast to number as well. Instead should we be casting to bigint somewhere in `tvlEvents.reduce((acc, event) => acc + event.amount, 0)`?

jophish: Hrm, yeah, that makes sense. I guess I was surprised that `getEvents` returns `amount` as a `Number`, since natively the type returned from fetching events would be a `bigint` - I would expect `getEvents` to return a raw `bigint` and allow callers to do what they want with it from there. Given the way that it works now, this makes sense!

jophish (approved):

PR #67, opened 2025-03-20
chore: add unit test for `RewardPool` contract

bakoushin:
Add unit test for `RewardPool` contract.

To facilitate ERC-20 tokens interaction, a mock ERC-20 token contract added to `contracts/mocks` directory.

Related to ENG-204


contracts/mocks/MockERC20.sol
```diff
@@ -0,0 +1,19 @@
...
+    string memory _symbol
+  ) ERC20(_name, _symbol) {}
+
+  function mint(address account, uint256 amount) external {
+    _mint(account, amount);
+  }
+
+  function burn(address account, uint256 amount) external {
+    _burn(account, amount);
+  }
```
finnian0826, line 18: Is this `burn` function needed? It doesn't seem to be used, and I don't see it in other mockERC20 implementations we have ([example](https://github.com/valora-inc/learn2earn-rewards/blob/main/contracts/mocks/MockErc20.sol)).

test/RewardPool.ts
```diff
@@ -0,0 +1,920 @@
...
+      mockERC20,
+      owner,
+      manager,
+      user1,
+      user2,
+      stranger,
+    }
+  }
+
+  async function deployNativeRewardPoolContract() {
```
finnian0826, line 67: This is very similar to the above helper function, could they be combined with a `isNative` param and just return an optional erc20?

test/RewardPool.ts
```diff
@@ -0,0 +1,920 @@
...
+          await rewardPool.DEFAULT_ADMIN_ROLE(),
+          owner.address,
+        ),
+      ).to.be.true
+      expect(
+        await rewardPool.hasRole(
+          await rewardPool.MANAGER_ROLE(),
+          manager.address,
+        ),
+      ).to.be.true
```
finnian0826, line 120: Should we also check that owner doesn't have "MANAGER_ROLE" and manager doesn't have "DEFAULT_ADMIN_ROLE" to test that it just doesn't always return true?

test/RewardPool.ts
```diff
@@ -0,0 +1,920 @@
...
+    const depositAmount = hre.ethers.parseEther('1')
+
+    describe('with ERC20 token', function () {
+      let rewardPool: Contract
+      let manager: HardhatEthersSigner
+      let stranger: HardhatEthersSigner
+      let pool: Contract
+
+      beforeEach(async function () {
+        const deployment = await loadFixture(deployERC20RewardPoolContract)
```
finnian0826, line 176: Do we need to deploy before each of the following or could all be done on one deployment?

test/RewardPool.ts
```diff
@@ -0,0 +1,920 @@
...
+        manager = deployment.manager
+
+        // Connect with manager
+        pool = rewardPool.connect(manager) as typeof rewardPool
+
+        // Deposit
+        await pool.deposit(nativeDepositAmount, { value: nativeDepositAmount })
+      })
+
+      it('allows manager to withdraw after timelock', async function () {
```
jeanregisser, line 348: I see the native case only tests for this single case, compared to the ERC20 block which covers more.

Should should make those tests more generic so we can `describe.each` using an ERC20 and native tokens and ensure we test the same cases?

jeanregisser (approved):
Awesome! 🚀

contracts/mocks/MockERC20.sol
bakoushin: yeah, this is redundant, it could be removed

test/RewardPool.ts
bakoushin: great catch! 

I think having the ERC20 deployed won't hurt the native test case but will make the boilerplate configuration simpler.

will use the single universal helper for all.

test/RewardPool.ts
bakoushin: I guess it is more predictable and robust when each case starts from the same clean state. it may be easy to overlook some issues when the later cases are influenced by the earlier ones.

test/RewardPool.ts
bakoushin: well, the `hasRole` method comes from the [openzeppelin library](https://docs.openzeppelin.com/contracts/3.x/api/access#AccessControl-hasRole-bytes32-address-), and we _expect_ it to be well-tested and to behave properly

test/RewardPool.ts
bakoushin: yes, that make sense to ensure we cover both token types.

refactored the token-related sections to loop over `ERC20` and `native`:
* Initialization
* Deposit
* Withdraw
* Claim reward
* Rescue tokens

PR #70, opened 2025-03-20
chore: configure deployment script for `RewardPool` contract

bakoushin:
Configure deployment script for `RewardPool` contract

To simplify a multi-contract deployment from a single repo, the deployment scripts migrated from a custom deployment script to [Hardhat Tasks](https://hardhat.org/hardhat-runner/docs/advanced/create-task#advanced-usage).

Related to ENG-204


package.json
```diff
@@ -61,6 +60,7 @@
     "jest": "^29.7.0",
     "knip": "^5.46.0",
     "nock": "^13.5.6",
+    "node-fetch": "^2.7.0",
```
jeanregisser, line 63: This shouldn't be needed, fetch is natively available since node 18.

hardhat.config.ts
```diff
@@ -15,6 +16,182 @@ const accounts: HDAccountsUserConfig = {
...
+  })
+
+task('deploy-mock-token', 'Deploy mock ERC-20 token').setAction(
+  async (_, hre) => {
+    await deployContract(hre, 'MockERC20', ['Mock ERC20', 'MOCK'])
+  },
+)
+
+// Contract deployment helper
+async function deployContract(
```
jeanregisser, line 134: Could we also print the command to verify the contract?
This ensures we don't forget doing it, since it can just be copy pasted from the output.

hardhat.config.ts
```diff
@@ -15,6 +16,182 @@ const accounts: HDAccountsUserConfig = {
...
+      })
+      await result.waitForDeployment()
+    } else {
+      await Contract.deploy(...constructorArgs)
+    }
+  }
+}
+
+// Contract upgrade helper
+async function upgradeContract(
```
jeanregisser, line 172: Ditto for the upragde.

hardhat.config.ts
jeanregisser, line 134: Could we print the constructor args too?
This helps ensure it's being deployed with the expected values.

hardhat.config.ts
```diff
@@ -15,6 +16,182 @@ const accounts: HDAccountsUserConfig = {
...
+  if (config.useDefender) {
+    console.log(`Upgrading ${contractName} with OpenZeppelin Defender`)
+    await hre.defender.proposeUpgradeWithApproval(proxyAddress, Contract, {
+      salt: config.defenderDeploySalt,
+    })
+  } else {
+    console.log(`Upgrading ${contractName} with local signer`)
+    const result = await hre.upgrades.upgradeProxy(proxyAddress, Contract)
+    await result.waitForDeployment()
+  }
```
jeanregisser, line 192: Could we print the `proxyAddress` too?
To help spot possible mistakes.

hardhat.config.ts
```diff
@@ -15,6 +16,182 @@ const accounts: HDAccountsUserConfig = {
...
+  'polygon',
+  'op',
+  'base',
+  'berachain',
+  'vana',
+]
+
+const ONE_DAY = 60 * 60 * 24
+
+task('deploy-registry', 'Deploy Registry contract')
```
jeanregisser, line 32: supernit: wondering if we should use this convention? for more visual grouping of tasks.
The hardhat task guide mentions this for subtasks: https://hardhat.org/hardhat-runner/docs/advanced/create-task#subtasks
But well I guess here it's not really a subtask. Anyway doesn't matter either way.

```suggestion
task('deploy:registry', 'Deploy Registry contract')
```

jeanregisser (approved):
Awesome! 🚀 
Also I think this fixes being unable to use the `--some-arg` argument we were declaring with yargs.

hardhat.config.ts
silasbw, line 32: nit: can you move these to a file in a `tasks/` subdir? We do that w/ other projects, like valora-punks (dubious reference, I know).

README.md
```diff
@@ -20,13 +20,12 @@ Run the localtest in one terminal:
...
+### Registry contract
+
+Deploy Registry:
 
 ```
-# Copy-paste environment definition
-SHELL=true yarn --silent hardhat --network hardhat run scripts/deploy.ts
-# or use eval
-eval `SHELL=true yarn --silent hardhat --network hardhat run scripts/deploy.ts`
+yarn hardhat deploy-registry --network localhost
```
silasbw, line 28: Dumb question: what's the best way to record the state (e.g., address) from this so that one can interact with it (e.g., call a function)?  That's what the SHELL=true was for.

silasbw (commented):
Can you split the script -> task refactor of deploying the registry into a PR that is separate from the new logic for deploying the RewardPool?

README.md
bakoushin: @silasbw
would it suffice/hurt if we just _always_ print the deployed address? do we need a flag for that?

we can just log some useful information, like @jeanregisser is suggesting. for instance: 
```
Proxy address: 0x...
Implementation address: 0x...

Owner: 0x...
Owner transfer delay: 0

Reward token: 0x...
Reward function ID: 0x...
Timelock: NNN
Manager: 0x...

To verify run:
npx hardhat verify --network celo 0x...
```

package.json
bakoushin: ah, yes, thanks for pointing this out!
will address this in subsequent PR, following @silasbw proposal to separate the refactoring from adding the new script.

PR #71, opened 2025-03-21
feat: filter for valid aerodrome referrals

MuckT:
### Description

Filters out valid aerodrome referrals. A valid referral is where both of the following statements are true:
- That the user has made at least one transaction on Aerodrome
- All transactions were made after the referral event.

### Testing
- [x] Unit tests added
- [x] Checked the results of getBlock with [Viem's Get Block Example](https://viem.sh/docs/actions/public/getBlockNumber.html#example) & `fetchReferrals.ts` with beefy.

### Related Issues

[ENG-194](https://linear.app/valora/issue/ENG-194/filter-for-valid-aerodrome-referrals)


scripts/protocolFilters/aerodrome.ts
```diff
@@ -1,7 +1,48 @@
...
 
-// TODO(ENG-194): Check that the user has made at least one transaction on Aerodrome,
-// and all transactions were made after the referral event.
 export async function filter(event: ReferralEvent): Promise<boolean> {
-  return !!event
+  const client = getHyperSyncClient(AERODROME_NETWORK_ID)
+  const query = {
+    transactions: [
+      { to: SUPPORTED_LIQUIDITY_POOL_ADDRESSES, from: [event.userAddress] },
+    ],
```
finnian0826, line 14: I don't think this will give us the info that we are looking for to determine if a user has interacted with Aerodrome in the way we care about. 

The interactions we are interested in is when a user does a swap, and the swap route that they execute is routed through one of the supported liquidity pools. We check this by looking at the `Swap` events emitted from the liquidity pool and check if the user address is the `recipient`. The user is not the one calling some transaction on the pool contract, I think that usually comes from some type of swap routers.

MuckT: Fixed in 67250922fa0d6c0a330c14e86c06ee5394a754e0.

scripts/protocolFilters/aerodrome.ts
```diff
@@ -1,7 +1,48 @@
...
+    fieldSelection: {
+      block: [BlockField.Number],
+    },
+    fromBlock: 0,
+  }
+
+  let hasMoreBlocks = true
+  let foundValidTransaction = false
+
+  while (hasMoreBlocks) {
```
jophish, line 24: This is the second place we're implementing hypersync pagination, along with in [networks.ts](https://github.com/divvi-xyz/divvi-protocol-v0/blob/main/scripts/calculateRevenue/protocols/utils/networks.ts#L5). I wonder if it's worth it to unify the hypersync requests/pagination into a single helper. If it were something simpler than pagination, I'd say it's fine to keep them duplicated, but since the pagination semantics are potentially a little subtle I wonder if it's worth combining these.

scripts/protocolFilters/aerodrome.ts
```diff
@@ -1,7 +1,48 @@
...
+    for (const block of response.data.blocks) {
+      if (block.number) {
+        const blockData = await getBlock(
+          AERODROME_NETWORK_ID,
+          BigInt(block.number),
+        )
+        if (blockData.timestamp < BigInt(event.timestamp)) {
+          return false // disqualify immediately if any tx is too early
+        } else {
+          foundValidTransaction = true
```
jophish, line 41: Are the results from hypersync guaranteed to be ordered? If so, can't we also just bail early and do a `return true` here?

scripts/protocolFilters/aerodrome.ts
MuckT: Adding in a separate PR: https://github.com/divvi-xyz/divvi-protocol-v0/pull/77.

scripts/protocolFilters/aerodrome.ts
MuckT: Changed in 0251df522c7aae3608bb41a5aa4f251cbca1af38!

scripts/protocolFilters/aerodrome.test.ts
```diff
@@ -0,0 +1,129 @@
...
+      .mockResolvedValueOnce(makeQueryResponse([]))
+      .mockResolvedValueOnce(makeQueryResponse([])) // causes loop to break
+
+    const result = await filter(event)
+    expect(result).toBe(false)
+    expect(mockClient.get).toHaveBeenCalledTimes(2)
+    expect(getBlock).not.toHaveBeenCalled()
+  })
+
+  it('handles paginated results and returns true if the first found block is after the referral event timestamp ', async () => {
```
finnian0826, line 94: Is this test needed since pagination was moved to a helper function in the other PR and has it's own unit tests now?

scripts/protocolFilters/aerodrome.test.ts
```diff
@@ -0,0 +1,129 @@
...
+  },
+  nextBlock,
+  totalExecutionTime: 50,
+})
+
+describe('filter', () => {
+  const userAddress = '0xUser'
+  const event: ReferralEvent = {
+    userAddress: userAddress,
+    timestamp: Math.round(new Date('2023-09-30T00:00:00Z').getTime() / 1000),
```
finnian0826, line 29: Timestamp is set here but then overwritten in each test, instead could we have it set here and just have the mocked `getBlock` return times such that the tests behave as expected?

finnian0826 (approved):
Some questions on tests but otherwise lgtm assuming that the pagination helper PR will also get merged in.

scripts/protocolFilters/aerodrome.test.ts
MuckT: Changed in 5c3468a0af5dea1f9121dfa6c2148abefa35763b!

scripts/protocolFilters/aerodrome.test.ts
MuckT: I'll look at removing in #77. I don't think doubling up has too much overhead in this case.

PR #72, opened 2025-03-21
chore: fix tsconfig ignoring tests

jeanregisser:
I noticed the test folder wasn't being typechecked.
Then I fixed the new issues reported.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -5,7 +5,7 @@ module.exports = {
   // Allow absolute imports from the tsconfig baseUrl
   moduleDirectories: ['node_modules', '<rootDir>'],
   transform: {
-    '^.+\\.(ts|tsx)$': ['ts-jest', { tsconfig: 'tsconfig.test.json' }],
+    '^.+\\.(ts|tsx)$': ['ts-jest', { tsconfig: 'tsconfig.json' }],
   },
   testPathIgnorePatterns: ['dist'],
   setupFiles: ['<rootDir>/jest.setup.js'],
--- a/package.json
+++ b/package.json
@@ -10,8 +10,7 @@
     "node": "^20"
   },
   "scripts": {
-    "build": "tsc",
-    "typecheck": "yarn build",
+    "typecheck": "tsc --noEmit",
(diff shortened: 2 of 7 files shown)


jest.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   // Allow absolute imports from the tsconfig baseUrl
   moduleDirectories: ['node_modules', '<rootDir>'],
   transform: {
-    '^.+\\.(ts|tsx)$': ['ts-jest', { tsconfig: 'tsconfig.test.json' }],
+    '^.+\\.(ts|tsx)$': ['ts-jest', { tsconfig: 'tsconfig.json' }],
```
jeanregisser, line 8: In this repo we can directly use the main tsconfig.

package.json
```diff
@@ -10,8 +10,7 @@
     "node": "^20"
   },
   "scripts": {
-    "build": "tsc",
```
jeanregisser, line 13: We don't need a TS build in this repo, we're no publishing anything.

scripts/setupTestnet.ts
jeanregisser, line 1: For some reason this started to be typechecked and complained.
Not really sure why it was previously ignored.

test/Registry.ts
```diff
@@ -140,14 +140,12 @@ describe(REGISTRY_CONTRACT_NAME, function () {
...
       await expect(
-        registry
-          .connect(addr1)
-          .registerReferrer(
-            mockReferrerIdHex,
-            protocolIds,
-            mockRewardRates,
-            mockRewardAddress,
-          ),
+        (registry.connect(addr1) as typeof registry).registerReferrer(
```
jeanregisser, line 143: Ideally we wouldn't need these casts.

However this is because we use `ethers` in those tests, because of `hardhat-upgrades` which uses it (see https://github.com/valora-inc/valora-points-contracts/pull/5#discussion_r1559275283, and https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/1071).

And it requires `typechain` to get proper types, which we could definitely do, but maybe there's a way to wrap with viem?

sviderock (approved):
🚀

PR #77, opened 2025-03-25
feat: hypersync pagination helper

MuckT:
### Description

This PR aims to create a shared helper for HyperSync pagination logic that logic to keep our code DRY.

### Testing
- [x] Unit tests added
- [x] Existing unit tests passing


scripts/protocolFilters/aerodrome.ts
```diff
@@ -1,47 +1,37 @@
...
-    const response: QueryResponse = await client.get(query)
-    if (response.nextBlock === query.fromBlock) {
-      hasMoreBlocks = false
-    } else {
-      query.fromBlock = response.nextBlock
-    }
+  let foundPriorTransaction = false
 
+  await paginateQuery(client, query, async (response) => {
     for (const block of response.data.blocks) {
```
finnian0826, line 23: These are in order right? So essentially we're just checking the first block that has a number and checking whether it is before or after the event timestamp

scripts/protocolFilters/aerodrome.ts
```diff
@@ -1,47 +1,37 @@
...
           AERODROME_NETWORK_ID,
           BigInt(block.number),
         )
-        if (blockData.timestamp < BigInt(event.timestamp)) {
-          return false // disqualify immediately if any tx is too early
-        } else {
-          return true // blocks are returned oldest first - qualify immediately if the block timestamp is after the event timestamp
-        }
+
+        foundPriorTransaction = blockData.timestamp >= BigInt(event.timestamp)
```
finnian0826, line 30: This isn't strictly `foundPriorTransaction`, it's more like "found a transactions after event.timestamp and didn't find any transactions before it"

finnian0826 (approved):
A question for my understanding and then variable should probably be renamed but otherwise lgtm!

scripts/protocolFilters/aerodrome.ts
MuckT: That's correct, and the blocks are in order.

scripts/protocolFilters/aerodrome.ts
MuckT: Changed to `hasTransactionsOnlyAfterEvent` in 782c1a0a74aac49aab4b40f994cd596ad2b985c4.

PR #78, opened 2025-03-25
feat(aerodrome): Compute trading fee for supported pools

finnian0826:
Just using the current fee as opposed to the fee at the time of the swap. Fee seems to be pretty constant, and there is no easy way to find historical value of fee (worked with Joe on this, I can provide more context if wanted).

It looks like only "concentrated" liquidity pools have trading fees (at least within the contract), so removed "basic" pools from the supported list.


scripts/calculateRevenue/protocols/aerodrome/constants.ts
```diff
@@ -3,14 +3,14 @@ import { NetworkId } from '../../../types'
...
   '0x70aCDF2Ad0bf2402C957154f944c19Ef4e1cbAE1', // WETH/cbBTC
-  '0xd9eDC75a3a797Ec92Ca370F19051BAbebfb2edEe', // WETH/KTA
   '0xC200F21EfE67c7F41B81A854c26F9cdA80593065', // VIRTUAL/WETH
   '0x6446021F4E396dA3df4235C62537431372195D38', // WETH/superOETHb
 ]
 export const AERODROME_NETWORK_ID = NetworkId['base-mainnet']
 
 export const TRANSACTION_VOLUME_USD_PRECISION = 8
+
+export const FEE_DECIMALS = 6
```
finnian0826, line 16: This is given on the fee read method of the [pool contract](https://basescan.org/address/0xb2cc224c1c9fee385f8ad6a55b4d94e92359dc59#readContract)

jh2oman (approved):

PR #79, opened 2025-03-27
feat(velodrome): Add velodrome calculation, make generic drome helper

finnian0826:
Just supporting one velodrome pool (WETH/USDC) for now


jophish (approved):
looks good!

PR #80, opened 2025-03-27
fix: fix deployment with owner other than deployer

bakoushin:
### Description

The Reward pool contract is unable to deploy with the Owner other than the deployer.

The problem is with granting Manager role during initialization. After the owner is set, the checks in `grantRole()` prevent deployer from granting the Manager role. This is inconvenient since we want to set both Owner and Manger during contract initialization.

This PR fixes it by replacing the `grantRole()` call during initialization with an internal `_grantRole()` to bypass role checks.

The tests, sadly, didn't cover the case to surface earlier. 

### Test plan

* Updated unit tests

### Related issues

Related to RET-204

### DIFF
--- a/contracts/RewardPool.sol
+++ b/contracts/RewardPool.sol
@@ -103,7 +103,7 @@ contract RewardPool is
     __UUPSUpgradeable_init();
     __ReentrancyGuard_init();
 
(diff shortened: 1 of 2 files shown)


jeanregisser (approved):
ah yes wasn't obvious to spot that one! 🚀

PR #81, opened 2025-03-28
chore(deploy): allow disable logging

bakoushin:
Allow to suppress all default deployment logging in Hardhat Tasks deployment helpers.

This will facilitate the migration of the Registry contract deployment to Hardhat Tasks. The Registry deployment may require printing shell commands instead of logs as described in the [README](https://github.com/divvi-xyz/divvi-protocol-v0#registry-contract).

Related to ENG-204

### DIFF
--- a/tasks/deployHelpers.ts
+++ b/tasks/deployHelpers.ts
@@ -41,13 +41,13 @@ export async function deployContract(
   let proxyAddress: string | undefined
   let contractAddress: string
 
-  console.log(`Deploying ${contractName}`)
-  console.log('Constructor args:', constructorArgs.join(', '))
+  log(`Deploying ${contractName}`)
+  log('Constructor args:', constructorArgs.join(', '))
 
   if (config.useDefender) {
-    console.log(`Using OpenZeppelin Defender`)
+    log(`Using OpenZeppelin Defender`)
     if (config.defenderDeploySalt) {
-      console.log(`Salt: ${config.defenderDeploySalt}`)
+      log(`Salt: ${config.defenderDeploySalt}`)
     }
 
     if (config.isUpgradeable) {
@@ -75,7 +75,7 @@ export async function deployContract(
       contractAddress = await contract.getAddress()
     }
   } else {
-    console.log(`Using local signer`)
+    log(`Using local signer`)
     if (config.isUpgradeable) {
       const proxy = await hre.upgrades.deployProxy(Contract, constructorArgs, {
         kind: 'uups',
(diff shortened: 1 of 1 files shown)


jeanregisser (commented):
Sorry, not fully following what `SHELL=true` means and what it allows.
Without knowing more it seems maybe we want a configurable log level or a `--silent` mode?

Could you help me understand?

bakoushin:
@jeanregisser 
great question! and yes, that is tricky!

what we want is to support the behavior described here (note the `eval` example):
https://github.com/divvi-xyz/divvi-protocol-v0#registry-contract

basically, when `SHELL=true`:
* suppress logging
* print some shell commands instead (e.g. set env variables)

I agree that in the helper function, we can have a logging parameter rather than reading the environment variable directly.
refactored it in that way!

see also 
* @silasbw comment: https://github.com/divvi-xyz/divvi-protocol-v0/pull/70#discussion_r2014645663
* draft PR for Registry deployment migration to Hardhat Tasks: #82

jeanregisser:
Thanks a lot for the context and links @bakoushin. Really helpful to see where the `SHELL=true` approach came from and what the original intent was. I appreciate you refactoring to make the logging configurable — that definitely helps make the behavior more predictable.

That said, I still find the `SHELL=true` pattern a bit surprising, especially since it's already been broken once. For example, here `console.log` ends up being called regardless of `SHELL`:  
https://github.com/divvi-xyz/divvi-protocol-v0/blob/40266521b478d543300ffc954c9cbcd2a0601691/scripts/deploy.ts#L102-L107

It's a clever idea, but it feels a bit brittle and easy to miss, especially for folks less familiar with the convention. I wonder if we might want to move toward something a little more standard and easier to maintain.

Personally, I’ve had good success with writing deployment info to a JSON file. We’ve done that before, like in this script:  
https://github.com/valora-inc/valora-eth-beta-nft/blob/c4d7a7b83b8127d5a75f5f450374b470042e9628/scripts/deploy.ts#L32-L35  
Or when using `hardhat-deploy`, which creates a structured JSON output by default.

That makes it easier for follow-up scripts to read the data consistently, whether they’re shell scripts, Hardhat tasks, or CI jobs, and avoids needing to eval shell output.

Curious what you think. I’m happy to help with a small migration PR if that feels like a good direction.

silasbw:
Thanks @jeanregisser -- fwiw, I think I'd prefer something like your custom deploy recording snippet instead of `hardhat-deploy`.  I found `hardhat-deploy` cumbersome to work w/ for relatively simple use cases.

bakoushin:
@jeanregisser I agree that having some artifacts exported to some file could be in _addition to_, but not _instead of_ the logs output.

happy to close this one, assuming we agree that we do not want to suppress logging for now.

as for the export format, I guess if the JSON is great if we want to consume it in JavaScript, maybe an `.env` file could serve both JavaScript and bash scripts (without jq).

PR #82, opened 2025-03-28
chore(deploy): migrate Registry contract deployment to Hardhat Tasks

bakoushin:
### Description

* Migrate Registry contract deployment to Hardhat Tasks
* Migrate Registry population with test data to Hardhat Tasks
* Write/read deployment metadata using the artifacts directory (alongside Hardhat artifacts)
  * For now, the metadata consists of the contract address only

Context: https://github.com/divvi-xyz/divvi-protocol-v0/pull/81#issuecomment-2761563357

### Related issues

Related to ENG-204


README.md
```diff
@@ -25,16 +25,13 @@ yarn hardhat node
 Deploy Registry:
 
 ```
-# Copy-paste environment definition
-SHELL=true yarn --silent hardhat --network hardhat run scripts/deploy.ts
-# or use eval
-eval `SHELL=true yarn --silent hardhat --network hardhat run scripts/deploy.ts`
+yarn hardhat registry:deploy --network localhost --output-file $TMPDIR/deployment-registry.json
```
jeanregisser, line 28: What do you think about writing the deployment info to a known location by default (in the repo, including the network name in the path)? Just thinking it could avoid needing to pass these flags around every time.

Then the populate task could read from the same default path — makes the whole flow easier to use out of the box, while still allowing overrides if needed.

Feels like a nice win for convention over configuration.

bakoushin:
@jeanregisser great idea, thanks! I made use of the `artifacts` directory Hardhat is using.
by default, the deployment metadata is always written there and can be read using the helper function.

there is also a helper for writing the metadata to this conventional location if any override is needed.

I think this has a good balance of relying on some stored output without requiring any extra configuration.

jeanregisser (approved):
Awesome work! 🚀

PR #83, opened 2025-03-28
chore: allow Owner to add rewards instead of Manager

bakoushin:
### Description

Allow Owner to add rewards instead of Manager.

We assume the Manager will be in control of the overall funds (can deposit and withdraw), but they won't assign individual rewards on a regular basis. 

We expect this will be done manually by our team for now. So we reserve this function for the contract owner.

### Test plan

* Updated unit tests

### Related issues

* Related to RET-204


jophish (commented):
Do we actually need to change the permissions for this function? Since the `DEFAULT_ADMIN_ROLE` is already the admin role for `MANAGER_ROLE`, any address with the `DEFAULT_ADMIN_ROLE` should be able to call methods that are restricted to `MANAGER_ROLE` -- see the paragraph after the code block [here](https://docs.openzeppelin.com/contracts/4.x/access-control#granting-and-revoking).

jeanregisser (commented):
Maybe dumb question, but do we really need a manager role?

bakoushin:
@jeanregisser we expect the Manager to be an external actor, likely representing some foundation.

we aim to grant them least privilege, limited to:
* depositing funds
* withdrawing funds

this also makes it clear that we don’t intend to manage the funds ourselves.

more context: https://docs.google.com/document/d/19u9zlMQ4cdoBRv34taxAYnGaLUJig6mfRDnbTboYwQM/edit?tab=t.0#heading=h.8t24a3up2325

jeanregisser:
> this also makes it clear that we don’t intend to manage the funds ourselves.

Yes though if we're owner for now, we could still do whatever we want with the funds. But I understand that's not the point for now.

jeanregisser (approved):

PR #87, opened 2025-04-01
chore: use contract:action naming convention in tasks

bakoushin:
Use the `contract:action` naming convention for all tasks, similar to the one implemented for the Registry task. 

It seems natural to put the contract name first, as we group tasks per contract, and different contracts may have different associated tasks.

This makes this naming approach consistent.

Resulting tasks:
* `registry:deploy`
* `registry:populate`
* `reward-pool:deploy`
* `reward-pool:upgrade`
* `mock-token:deploy`

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,13 +39,13 @@ yarn hardhat --network localhost registry:populate
 Deploy mock token:
 
 ```bash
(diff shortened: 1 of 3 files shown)


jeanregisser (approved):
🚀

PR #88, opened 2025-04-04
feat(fonbnk): Add revenue calculation for Fonbnk

finnian0826:
Uses hypersync to find all transfer events from the Fonbnk payout wallets to the user. Uses the "Transfer" event topic string `0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef`.

Tested manually on a user who has received funds from [one of the payout addresses](https://celoscan.io/address/0xd352dee5b113a6acd09e10eb10a1d8e6d91e615d), behaved as expected.

[Linear](https://linear.app/valora/issue/ENG-177/transaction-volume-for-fonbnk)


scripts/calculateRevenue/protocols/fonbnk/constants.ts
```diff
@@ -0,0 +1,26 @@
+import { NetworkId } from '../../../types'
+
+export const TRANSACTION_VOLUME_USD_PRECISION = 8
+
+export const FONBNK_API_URL = 'https://aten.fonbnk-services.com'
```
finnian0826, line 5: Should this be an environment variable? The beefy API URL is not.

scripts/calculateRevenue/protocols/fonbnk/index.ts
```diff
@@ -0,0 +1,155 @@
...
+  startTimestamp: Date
+  endTimestamp: Date
+}): Promise<number> {
+  if (!isAddress(address)) {
+    throw new Error('Invalid address')
+  }
+  let totalRevenue = 0
+  const fonbnkAssets = await getFonbnkAssets()
+  for (const supportedNetwork of Object.values(FonbnkNetwork)) {
+    const checkedAddresses = new Set<Address>()
```
finnian0826, line 120: The same payout address is used for multiple tokens on the same network, using this to not double count.

scripts/calculateRevenue/protocols/fonbnk/index.test.ts
```diff
@@ -0,0 +1,177 @@
...
+]
+
+const MOCK_ADDRESS = '0x1234567890123456789012345678901234567890' as Address
+
+describe('getUserTransactions', () => {
+  let mockClient: { get: jest.Mock }
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+  it('should fetch user transactions', async () => {
```
satish-ravi, line 80: same style comment as other files, consider grouping tests into sections, e.g.,
Arrange (mocks), Act, Assert.

scripts/calculateRevenue/protocols/fonbnk/constants.ts
finnian0826: It shouldn't change so should be okay as is.

scripts/calculateRevenue/protocols/fonbnk/helpers.ts
```diff
@@ -0,0 +1,95 @@
...
+import { FonbnkAsset, FonbnkPayoutWalletReponse } from './types'
+import { Address } from 'viem'
+
+export async function getFonbnkAssets(): Promise<FonbnkAsset[]> {
+  if (!process.env.FONBNK_CLIENT_ID) {
+    throw new Error('FONBNK_CLIENT_ID is not set')
+  }
+  if (!process.env.FONBNK_CLIENT_SECRET) {
+    throw new Error('FONBNK_CLIENT_SECRET is not set')
+  }
```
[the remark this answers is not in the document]
finnian0826: Would I add a parseArgs function to this file and then use that within these functions? Or maybe in the fonbnk `calculateRevenue` function so I only have to do it once? It doesn't seem like I can add it to the parseArgs function in the main `calculateRevenue` file because then I would have to pass all the values around.

scripts/calculateRevenue/protocols/fonbnk/index.ts
```diff
@@ -0,0 +1,155 @@
...
+  startTimestamp,
+  endTimestamp,
+  client,
+  networkId,
+}: {
+  address: Address
+  payoutWallet: Address
+  startTimestamp: Date
+  endTimestamp: Date
+  client: { get: (query: Query) => Promise<QueryResponse> }
```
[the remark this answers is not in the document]
finnian0826: This should work, but makes mocking harder in the tests. In all the other places hypersync is used it is like my initial implementation.

scripts/calculateRevenue/protocols/fonbnk/constants.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const TRANSFER_TOPIC = encodeEventTopics({
+  abi: erc20Abi,
+  eventName: 'Transfer',
+})[0]
+// Gets transfer topic hash, should be 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef
```
satish-ravi, line 13:
```suggestion
// Gets transfer topic hash, should be 0xddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef
export const TRANSFER_TOPIC = encodeEventTopics({
  abi: erc20Abi,
  eventName: 'Transfer',
})[0]
```

scripts/calculateRevenue/protocols/fonbnk/helpers.ts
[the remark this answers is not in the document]
satish-ravi: I think ideally this goes in the main parseArgs, so running `calculateRevenue -h` lists all the required params. We can also enforce strict checks with the `check` option and ensure if they are set when protocol is `fonbnk`. We can inject a `FonbnkConfig` type so we don't pass around too many args. If this is getting hard to do (>30 mins of your time), I'd suggest doing it locally here, can just be a simple function that reads process.env returns an object containing id and secret so that process.env access is limited to it.

scripts/calculateRevenue/protocols/fonbnk/index.ts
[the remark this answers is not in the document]
satish-ravi: I see, is the mocking an issue because it would require mocking all the methods in HypersyncClient ? If so, think we can avoid it with using `as any` or `as unknown`, etc in tests which we already do for other mocks. Another work around option is to use `Pick<HypersyncClient, 'get'>` so we don't have this explicitly defined in many places.

scripts/calculateRevenue/protocols/fonbnk/index.ts
```diff
@@ -0,0 +1,167 @@
...
+      log: [LogField.BlockNumber, LogField.Address, LogField.Data],
+    },
+    fromBlock: 0,
+  }
+
+  const transactions: FonbnkTransaction[] = []
+  await paginateQuery(client, query, async (response) => {
+    for (const transaction of response.data.logs) {
+      // Check that the logs contain all necessary fields
+      if (transaction.blockNumber && transaction.data && transaction.address) {
```
satish-ravi, line 45: is it bad if these fields are missing? Should we log if that happens?

scripts/calculateRevenue/protocols/fonbnk/index.test.ts
```diff
@@ -0,0 +1,200 @@
...
+      address: MOCK_ADDRESS,
+      startTimestamp: new Date('2025-01-01T00:00:00Z'),
+      endTimestamp: new Date('2025-01-03T00:00:00Z'),
+    })
+
+    expect(getFonbnkAssets).toHaveBeenCalledTimes(1)
+    expect(getPayoutWallets).toHaveBeenCalledTimes(2)
+    // The first included transaction has hex value 0x2710 with 4 decimals which is 1, with a price of 3 that is 3 USD
+    // The second included transaction has hex value 0x88B8 with 4 decimals which is 3.5, with a price of 5 that is 17.5 USD
+    expect(result).toEqual(41)
```
satish-ravi, line 198: how is 41 arrived from 3 and 17.5?

scripts/calculateRevenue/protocols/fonbnk/index.ts
finnian0826: If these fields are missing it means we can't get the timestamp (to check if in window), the amount, or the token transferred respectively. The types for them are all optional, but I'm not sure how often they would actually not be set? I can add some logging in the case that it does happen.

scripts/calculateRevenue/protocols/fonbnk/index.test.ts
finnian0826: comment added, it is because there are 2 payout addresses so each is counted twice (because of how the mocking is done)

scripts/calculateRevenue/protocols/fonbnk/helpers.ts
```diff
@@ -4,25 +4,33 @@ import { FonbnkAsset, FonbnkNetwork, FonbnkPayoutWalletReponse } from './types'
 import { Address } from 'viem'
 import { fetchWithBackoff } from '../../../utils/fetchWithBackoff'
 
-export async function getFonbnkAssets(): Promise<FonbnkAsset[]> {
+function getFonbnkEvnVariables() {
```
satish-ravi, line 7:
```suggestion
function getFonbnkEnvVariables() {
```
or 
```suggestion
function getFonbnkConfig() {
```

scripts/calculateRevenue/protocols/fonbnk/index.ts
```diff
@@ -56,6 +56,10 @@ export async function getUserTransactions({
             timestamp: blockTimestampDate,
           })
         }
+      } else {
+        console.log(
+          `Fonbnk transfer transaction missing one of the required fields: ${transaction}`,
```
satish-ravi, line 61: I don't think this will serialize correctly. IIRC this might just print `[object]` or something similar.

scripts/calculateRevenue/protocols/fonbnk/index.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    amount: BigInt(35000),
+    tokenAddress: '0x123',
+    timestamp: new Date('2025-01-02T21:30:00.000Z'),
+  },
+]
+
+const MOCK_ADDRESS = '0x1234567890123456789012345678901234567890' as Address
+
+describe('getUserTransactions', () => {
+  let mockClient: { get: jest.Mock }
```
satish-ravi, line 95:
```suggestion
  const mockClient: HypersyncClient = { get: jest.fn() } as unknown as HypersyncClient
```

scripts/calculateRevenue/protocols/fonbnk/index.test.ts
```diff
@@ -0,0 +1,205 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+  it('should fetch user transactions', async () => {
+    mockClient = { get: jest.fn() }
```
satish-ravi, line 100:
```suggestion
```

scripts/calculateRevenue/protocols/fonbnk/index.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    jest
+      .mocked(getHyperSyncClient)
+      .mockReturnValue(
+        mockClient as unknown as ReturnType<typeof getHyperSyncClient>,
+      )
```
satish-ravi, line 105:
```suggestion
      .mockReturnValue(mockClient)
```

scripts/calculateRevenue/protocols/fonbnk/index.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    mockClient.get
```
satish-ravi, line 106:
```suggestion
    jest.mocked(mockClient.get)
```

satish-ravi (approved):

PR #91, opened 2025-04-10
feat: add DivviRegistry contract with entities and agreements handling

kathaypacific:
~~I followed the guide [here](https://docs.google.com/document/d/12HZQ-e5mWM_oAhKosKg86TdQp0uUPlCZbwI-F-MVoz4/edit?tab=t.0) for the expected contract methods. ~~

~~As discussed, I've omitted the remove referral / remove agreement methods for now since they make our lives complicated. Owner transferability was also [removed](https://valora-app.slack.com/archives/C08EUKGJP4H/p1744379049817239) in favour of immutable owner / entity id.~~

~~The contract does not keep state that facilitates getting lists of referrals to each consumer / provider, it's intended that information like that should be retrieved from emitted events.~~

The scope of this PR has been updated to:
- Add the new upgradable version of the DivviRegistry contract
- Add storage and methods that facilitate registering as an entity
- Add storage and methods that facilitate creating agreements and requiring approvals
- Tests for the above contract functionality

Will be addressed in follow ups:
- user referral management
- deploy and upgrade scripts

Assumptions:
1. Only providers require approval for new agreements. Consumers don’t need to approve agreements initiated by providers because:
- They’re likely to want the additional rewards.
- They can simply ignore a provider’s rewards contract if uninterested.
2. Consumers are expected to initiate contact with providers that require approval (e.g. via email, Discord). Approval decisions may depend on off-chain context that can’t be captured in a contract call / without some form of human interaction.

Flows:
Consumer calls `registerAgreementAsConsumer`
→ Succeeds (agreement created) or reverts.
→ If reverted with `ProviderRequiresApproval`, consumer reaches out to provider.
Provider calls `registerAgreementAsProvider`
→ Always succeeds if consumer exists.
→ If contacted by a consumer, the provider knows the address to use.

Related to ENG-234, ENG-298


contracts/RegistryV1.sol
```diff
@@ -0,0 +1,300 @@
...
+  function getRequiresApprovalForRewardsAgreements(
+    address entity
+  ) external view returns (bool) {
+    return _requiresApproval[_entityToId[entity]];
+  }
+
+  /**
+   * @notice Set whether a Rewards Entity requires approval for agreements
+   */
+  function setRequiresApprovalForRewardsAgreements(
```
kathaypacific, line 154: in the design doc we specified only `requireApprovalForRewardsAgreements` with no way to undo this action, but i implemented a toggle-able function here assuming that providers may regret this decision (since they create extra work for themselves, that they may not realise)

contracts/RegistryV1.sol
```diff
@@ -0,0 +1,313 @@
...
+    return true;
+  }
+
+  /**
+   * @notice Get the consumer address for a given user and provider
+   * @param user The address of the user
+   * @param provider The address of the provider
+   * @return The address of the consumer
+   */
+  function getReferringConsumer(
```
kathaypacific, line 286: i'm not sure if we need this, but in case a registerReferral is unsuccessful and the event emitted is "UserAlreadyReferred", i thought a builder might want an easy way to know which consumer the user is referred by for debugging?

contracts/RegistryV1.sol
```diff
@@ -0,0 +1,313 @@
...
+   * @return success Whether the referral was successfully registered
+   */
+  function registerReferral(
+    address user,
+    bytes memory userSignature,
+    address rewardsConsumer,
+    address rewardsProvider
+  ) public returns (bool success) {
+    // Verify user signature
+    bytes32 hash = keccak256(abi.encodePacked('Confirm referral'));
```
kathaypacific, line 226: magic string that consumers will need to bake into their apps....is there a better way to verify a signature without being opinionated about the signed message?

contracts/RegistryV1.sol
```diff
@@ -0,0 +1,313 @@
...
+    address user;
+    bytes userSignature;
+    address rewardsConsumer;
+    address rewardsProvider;
+  }
+
+  // Storage
+  uint256 private _nextEntityId; // Counter for generating unique id's
+  mapping(uint256 => address) private _idToEntity; // id => current owner address (Rewards Entity) allows each entity to have a stable identifier while allowing the owner to change
+  mapping(address => uint256) private _entityToId; // current owner => id
```
kathaypacific, line 32: the id management is a bit crazy, but given that the entities are identified by the owner address and we are allowing the transfer of that owner address, i opted to go for this option where the contract knows entities by an internal id that never changes

contracts/RegistryV1.sol
```diff
@@ -0,0 +1,313 @@
...
+
+  /**
+   * @notice Registers a user as being referred to a rewards agreement
+   * @param user The address of the user being referred
+   * @param userSignature The user's signature authorizing the referral
+   * @param rewardsConsumer The address of the rewards consumer entity
+   * @param rewardsProvider The address of the rewards provider entity
+   * @return success Whether the referral was successfully registered
+   */
+  function registerReferral(
```
jeanregisser, line 219: I'm evolving my thoughts about the need for batch processing referrals.
If we don't do batch processing, we could go back to normal revert reasons and make this function's behavior more standard/predictable.

I'll post more on this on Slack relayed to the research on relayers.

jeanregisser (commented):
Great work! 🙌

contracts/RegistryV1.sol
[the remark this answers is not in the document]
kathaypacific: yes all of this complexity is to enable `transferRewardsEntityOwner` - i don't have a lot of context about why it is needed, i was thinking that these entity addresses are also the wallets that will be used to distribute / claim rewards so perhaps there is a need for entities to be able to change it? @jophish might have more context to share. i'm happy to remove `transferRewardsEntityOwner` and simplify the state if it's not needed!

contracts/RegistryV1.sol
kathaypacific: yes that's fair, although do you think the message should contain `rewardsProvider`? it's more accurate but would require users to sign more than 1 message if the consumer app supports more than 1 provider

jeanregisser: Oh good question, to which I have another question, do we even need to pass the `rewardsProvider` for the referral call?
We could say, if the rewards consumer is referring a user, they automatically are linked to all providers for which the rewards consumer has an agreement with.
This saves on unnecessary additional calls. As more providers get added to the agreements from a  rewards consumer.
Would this make sense?

kathaypacific: yes i wondered about this too, i initially started with this approach but i ran into:
1. data storage complexities - registering a user against all agreements that the consumer has requires us to keep a mapping of consumer -> agreements to find the agreements
2. i couldn't think through what should happen if some of the referrals fail. e.g. a consumer has 3 agreements, tries to refer user X, but user X has already been referred to one of the providers by another app
3. similar to 2, i also couldn't think through how a consumer would handle referring users to a _new_ agreement if those users have already been referred to their existing agreements. if the number of referrals that the transaction should make is abstracted away, we would need to monitor transaction events to understand which ones succeeded and which ones failed?

i guess i kind of landed on one referral to one agreement per transaction as the simplest and most transparent approach, even if it is less efficient due to needing more transactions?

contracts/RegistryV1.sol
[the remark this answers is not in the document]
jeanregisser: I was just thinking, another reason to make entities addresses immutable: if apps hardcode them for the referrals (like we're doing with protocol IDs in v0), they must not change.
Or we would need to expose the internal IDs, or keep track of old addresses to still allow referrals with older addresses.
Right?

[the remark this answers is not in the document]
kathaypacific: yes that's a really good point. i think that in general identifiers should be unique and stable, and all this complexity is surfacing because we are trying to make an entity identifier mutable

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,347 @@
...
+  uint256 private _nextEntityId; // Counter for generating unique id's
+  mapping(uint256 => address) private _idToEntity; // id => current owner address (Rewards Entity) allows each entity to have a stable identifier while allowing the owner to change
+  mapping(address => uint256) private _entityToId; // current owner => id
+
+  // Agreement storage
+  mapping(bytes32 => bool) private _agreements; // keccak256(providerId, consumerId) => true (if agreement exists)
+  mapping(uint256 => bool) private _requiresApproval; // entityId => boolean (if entity requires approval)
+
+  // Referral tracking
+  mapping(address => mapping(uint256 => uint256)) private _userReferrals; // user => providerId => consumerId
```
bakoushin, line 39: what's the reason to make these private?

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,347 @@
...
+
+    if (_entityToId[entity] != 0) {
+      revert EntityAlreadyExists(entity);
+    }
+
+    uint256 entityId = ++_nextEntityId;
+    _idToEntity[entityId] = entity;
+    _entityToId[entity] = entityId;
+
+    emit RewardsEntityRegistered(entity);
```
bakoushin, line 112: not sure, but maybe also emit the `entityId` for completeness?

contracts/DivviRegistry.sol
kathaypacific: i was thinking of these entity id's as an internal contract state. officially, the requirement is that entities are identified by the owner address. but this can change, and i did not want to have to update all the agreements / existing referrals when this happens (seems gas inefficient). so there is a need to be able to identify an entity by a stable id, but it is randomly assigned at registration time and does not mean anything to anyone outside of the contract. that is why i did not emit the id's. these mappings don't need to be private i suppose, but they are not designed to be readable. what is the benefit of making these public?

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+   * @dev Requires REFERRAL_REGISTRAR_ROLE
+   * @param user The address of the user being referred
+   * @param rewardsConsumer The address of the rewards consumer entity
+   * @param rewardsProvider The address of the rewards provider entity
+   */
+  function registerReferral(
+    address user,
+    address rewardsConsumer,
+    address rewardsProvider
+  ) external entityExists(rewardsProvider) entityExists(rewardsConsumer) {
```
jophish, line 171: Can we use the OpenZeppelin `onlyRole` modifier here? (see [here](https://docs.openzeppelin.com/contracts/4.x/access-control#using-access-control)) Or are we not using it explicitly so that we can emit a custom revert reason?

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+    _requiresApproval[msg.sender] = requiresApproval;
+    emit RequiresApprovalForRewardsAgreements(msg.sender, requiresApproval);
+  }
+
+  /**
+   * @notice Register a Rewards Consumer - Rewards Provider relationship
+   * @dev Should be called by the Rewards Consumer
+   * @param rewardsProvider The provider entity address
+   */
+  function registerRewardsAgreement(
```
jophish, line 124: If a provider requires approvals, how is the consumer supposed to ever register the agreement? I don't see a way in the contract for e.g., a consumer to "propose" an agreement with a provider that requires approvals, such that the provider can later approve it.

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+    _agreements[agreementKey] = true;
+    emit RewardsAgreementRegistered(rewardsProvider, msg.sender);
+  }
+
+  /**
+   * @notice Approve a rewards agreement
+   * @dev Should be called by the Rewards Provider
+   * @param rewardsConsumer The consumer entity address
+   */
+  function approveRewardsAgreement(
```
jophish, line 149: Ah, I see this re: my comment above. This feels slightly weird to me though, since when a consumer attempts to register an agreement with a provider that requires approvals, nothing is updated in contract state to reflect that there was a request for an agreement. There won't even be an emitted event that can be looked up. It seems like there should be a way for providers to check "pending" agreements that they can later approve or deny.

Further, the naming of this function actually seems a bit misleading; it's not approving a proposed reward agreement so much as it's allowing a provider to create a rewards agreement with any arbitrary consumer without the consumer's consent. It also doesn't require that the provider calling the method actually requires approvals, which feels a bit strange.

jophish: It also bypasses the check for whether the agreement key already exists, which should in theory revert with `AgreementAlreadyExists` as above I would imagine.

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,244 @@
...
+  modifier entityExists(address entity) {
+    if (!_entities[entity]) {
+      revert EntityDoesNotExist(entity);
+    }
+    _;
+  }
+
+  /**
+   * @notice Register a new rewards entity
+   * @param entity The entity address to register
```
jophish, line 88: Can the docstring for `entity` be expanded a bit to explain that the `entity` address should be owned by the entity and will be used to call privileged methods on the contract? Maybe a bit pedantic but it might not be immediately obvious what the purpose of `entity` is; I think it could be natural to assume that `msg.sender` is somehow important here, but it doesn't seem to be.

jophish: Actually, there could be a (very minor) pseudo-security annoyance with this; if you don't need to somehow prove ownership of the address you're setting for `entity` (e.g., by having `entity == msg.sender` implicitly), you can just chuck in whatever address you want and kinda address-squat on existing addresses. For example I could find the operational wallet addresses of a bunch of existing protocols and just register entities against those addresses - which wouldn't be the end of the world, but probably very annoying.

contracts/DivviRegistry.sol
kathaypacific: i guess implicitly i also assume that consumers would never turn down an agreement with a provider (because why would they not want the extra rewards) and therefore it is always possible for a provider to register an agreement with a consumer without their consent / initiation

contracts/DivviRegistry.sol
kathaypacific: i didn't check for `msg.sender` here because i think we are planning to register on behalf of some entities. i could restrict this method to `msg.sender` and the some other role (not the owner, because it's annoying to wait for extra signatures) only to be extra cautious? i don't see why people would pay gas to register addresses that they do not own though, it doesn't seem like there is any benefit for them and does not prevent the owners of those addresses to be properly integrated with divvi...i guess there is the scenario where some unrelated person who would register some addresses that belong to a well known protocol, leading builders to see those addresses and annoy the protocol by reaching out to them for agreements but i think this risk would be offset but the lack of advertising of the entity id anywhere :/

contracts/DivviRegistry.sol
kathaypacific: if we do want a pending agreements mapping on-chain, `registerRewardsAgreement` then would either create an agreement _or_ create a pending agreement right...? the multiple outcomes of this method would be confusing? i prefer the simplicity of:
1. consumer calls `registerRewardsAgreement` -> it either succeeds, or reverts. if it reverts because the provider wants to do manual approvals, the consumer needs the provider to create the agreement. 
2. provider creates the agreement (no checks necessary aside from the consumer exists, assuming consumers would never turn down extra rewards / if they don't want to engage with the provider, they could just not engage with that provider's reward contract)

(sorry for the wall of text)

jeanregisser: Thanks for detailing this. Makes sense to me too 👍

contracts/DivviRegistry.sol
jeanregisser: I know the likelihood of someone actually paying gas just to register bogus or other entities’ addresses is low, but I’d still prefer to avoid any confusion around “who owns what.” Enforcing a straightforward proof of ownership - like making `registerRewardsEntity` require `entity == msg.sender` or a signature from that address - would keep things truly self-service and eliminate any confusion.

For initial testing (e.g., registering Beefy, etc), I was thinking of creating a Safe on their behalf (tagged “ValoraBeefy” or similar), and later transferring ownership to them. They could also register a new entity themselves whenever they’re ready.

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,189 @@
...
+  Initializable,
+  AccessControlDefaultAdminRulesUpgradeable,
+  UUPSUpgradeable
+{
+  // Entities storage
+  mapping(address => bool) private _entities; // entity => true (if entity exists)
+
+  // Agreement storage
+  mapping(bytes32 => bool) private _agreements; // keccak256(provider, consumer) => true (if agreement exists)
+  mapping(address => bool) private _requiresApproval; // entity => boolean (if entity requires approval)
```
jeanregisser, line 22: How about grouping `exists` and `requiresApproval` into an `EntityData` struct?

This could make the entity data a bit more organized and potentially saves some gas due to storage packing when accessing both flags.

```suggestion
  struct EntityData {
      bool exists;
      bool requiresApproval;
      // fields can be added here in a future upgrade if needed
      // this is upgrade safe as long as `EntityData` is only used in a mapping
}

  // Entities storage
  mapping(address => EntityData) private _entities;
  // Agreement storage
  mapping(bytes32 => bool) private _agreements; // keccak256(provider, consumer) => true (if agreement exists)
```

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,189 @@
...
+    emit RewardsAgreementRegistered(msg.sender, rewardsConsumer);
+  }
+
+  /**
+   * @notice Check if an agreement exists between a consumer and provider
+   * @param provider The provider entity address
+   * @param consumer The consumer entity address
+   * @return exists Whether the agreement exists
+   */
+  function agreementExists(
```
jeanregisser, line 160: supernit: follows what OZ uses, like `hasRole`.
```suggestion
  function hasAgreement(
```

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,189 @@
...
+  }
+
+  /**
+   * @notice Register a Rewards Consumer - Rewards Provider relationship
+   * @dev Should be called by the Rewards Consumer
+   * @param rewardsProvider The provider entity address
+   */
+  function registerAgreementAsConsumer(
+    address rewardsProvider
+  ) external entityExists(rewardsProvider) entityExists(msg.sender) {
```
jeanregisser, line 115: Quick thought on agreement registration:

The current design doesn't prevent registering an agreement between, say, two entities that might conceptually only be "Consumers".

Is the intention that any registered entity can act as either a Provider or Consumer in an agreement, offering maximum flexibility? Just confirming this flexibility is desired over stricter role enforcement during agreement creation.

kathaypacific: yes it was documented [here](https://docs.google.com/document/d/12HZQ-e5mWM_oAhKosKg86TdQp0uUPlCZbwI-F-MVoz4/edit?tab=t.0), Joe gave the example where beefy might be a provider and a consumer, where aave might be it's provider. are there downsides to this flexibility? i haven't thought about this a ton but it seems nice that an entity can register once and be both (otherwise they would need to maintain two entities to be both?)

jeanregisser: Thanks yes makes sense, I like the flexibility.

contracts/DivviRegistry.sol
kathaypacific: makes sense, updated!

contracts/DivviRegistry.sol
kathaypacific: yes that seems more neat, updated!

contracts/DivviRegistry.sol
kathaypacific: yes makes sense, i've added a check for msg.sender now

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,197 @@
...
+  /**
+   * @notice Register a new rewards entity
+   * @param entity An address that the entity owns, which will be used to identify the entity and call privileged functions on this contract
+   * @param requiresApproval Whether the entity requires approval for agreements
+   */
+  function registerRewardsEntity(
+    address entity,
+    bool requiresApproval
+  ) external {
+    if (msg.sender != entity) {
```
jophish, line 90: `entity` seems like a redundant parameter now, since we can always just pull the entity address from `msg.sender`, since they must always be equal anyways in order for this function to do anything.

contracts/DivviRegistry.sol
```diff
@@ -0,0 +1,188 @@
...
+    }
+    _;
+  }
+
+  /**
+   * @notice Register the caller as a new rewards entity
+   * @param requiresApproval Whether the entity requires approval for agreements
+   */
+  function registerRewardsEntity(bool requiresApproval) external {
+    if (_entities[msg.sender].exists) {
```
jh2oman, line 85: For `registerRewardsEntity` and `registerAgreementAsConsumer` can we utilize a signature instead of using msg.sender to get the entity's address?

This would allow us to build gas-less registration of entities which could be useful for an Ambassador program or an invite rewards scenario where the RewardsConsumer is just a person as opposed to an app.

cc @silasbw

kathaypacific: that has tradeoffs for the usability of these 2 methods though, because the signed message would need to be generated first and then passed as a method argument so entities can't just do the transaction straight from a block explorer / Safe UI? it seems like we should create new methods for those and restrict the caller to a known address instead?

kathaypacific: i propose that we address this requirement in a follow up PR (this is a new requirement that i am just hearing about now). this contract won't be deployed until we select a home network so there is no redeployment overhead to address this later

jeanregisser: Yes we can punt on this.

An an alternative to custom signed messages, we could adopt ERC-2771 for this.
This is what Gelato recommends: https://docs.gelato.network/web3-services/relay/erc-2771-recommended/sponsoredcallerc2771
We'd still need signing from the user and trusted relayer(s), but it is standard.

jeanregisser (approved):
🚀

contracts/DivviRegistry.sol
jh2oman: okay, sounds good. Thanks 👍

jh2oman: created this ticket https://linear.app/valora/issue/ENG-309/divviregistry-allow-registering-entities-on-their-behalf

PR #93, opened 2025-04-10
feat(aave): aave revenue calculation

bakoushin:
### Description

Calculate Aave revenue in USD for a list of users during specific period of time.

### TLDR

1. Aave already has a logic for sharing its revenue with the LPs.
2. Let's derive the referrer share from the LP share.
3. Let's use the Reserve Factor for that.

### Protocol revenue calculation

Protocol revenue is estimated by multiplying user earnings by the reserve ratio — the portion of interest the protocol keeps before distributing the remainder to users.

We assume that:
* user earnings come from `1 - reserveFactor` of total interest
* protocol revenue come from `reserveFactor` of total interest
* therefore the ratio of protocol to user earnings is `reserveFactor / (1 - reserveFactor)`

> **Reserve Factor** A percentage of interest accrued by borrowers that is allocated to the Aave Treasury to help safeguard the protocol. Current reserve factors: https://aave.com/docs/resources/parameters

### User earnings calculation

To calculate user earnings accurately, we track all balance changes during the period. Earnings are measured as the change in the liquidity index during segments where the user’s balance remained unchanged.

> **Liquidity Index** Tracks the cumulative interest earned by a reserve over time, used to calculate accurate interest payments.

### USD conversion

Protocol revenue across all reserves is converted to USD using Aave oracle prices from the final block of the requested time period.

### Usage example

Required API keys:
* Alchemy
* Hypersync
* TheGraph (new one, [how to get an api key](https://thegraph.com/docs/en/subgraphs/querying/managing-api-keys/#create-and-manage-api-keys))

```bash
# create referrals list
echo "undefined,0xbe51b497abddd241aa9325021ec225cdfc9a9e48,1718386363" > aave_referrals.csv

# run the calculation script
yarn ts-node ./scripts/calculateRevenue.ts --protocol aave --startTimestamp 1718386363000 --endTimestamp 1741899467000

# print the results
cat aave_revenue.csv

```

### Test plan
 
Added unit test

### Related issues

Fixes ENG-251


scripts/calculateRevenue/protocols/aave/index.ts
```diff
@@ -41,8 +66,202 @@ export async function revenueInNetwork(
...
+      revenueInReserve(
+        reserveTokenAddress,
+        reserveTokenDecimals,
+        aTokenAddress,
+        context,
+      ),
+  )
+}
+
+export function revenueInReserve(
```
jeanregisser, line 94: nit: are we only exporting this because of unit tests? I understand it's convenient, but it's usually preferable to test only the public interface of a module.

scripts/calculateRevenue/protocols/aave/index.ts
```diff
@@ -41,8 +66,202 @@ export async function revenueInNetwork(
...
+  return userEarningsInSegment
+}
+
+// Calculates protocol revenue based on user earnings and reserve factor.
+// The calculation uses the relationship between user earnings and protocol revenue:
+// - User earnings come from (1 - reserveFactor) of total interest
+// - Protocol earnings come from (reserveFactor) of total interest
+// - Therefore the ratio of protocol to user earnings is: reserveFactor / (1 - reserveFactor)
+//
+function estimateProtocolRevenue(
```
jeanregisser, line 227: I'm having a hard time understanding that.
Is there some more docs I could read?
And why is this an estimate?

jeanregisser (commented):
Thanks for putting this together! 🚀 

The overall approach of using segmentation based on on-chain data seems logical for calculating user-attributed Aave revenue.

Since I'm not intimately familiar with Aave's internal accounting, I can't fully verify the accuracy. 

How can we best gain confidence in this calculation? Perhaps through specific examples or additional review?

scripts/calculateRevenue/protocols/aave/index.ts
bakoushin: I had a hard time too trying to find explicit docs on how Aave revenue is generated.
Perhaps due to my poor googling skills, perhaps because it is not explicitly described.

I ended up looking into their smart contracts to figure out what they do inside, as an ultimate source of truth.

#### Aave revenue: total amount

It seems they have 4 major sources of funds they put in their pocket (aka the "Collector Contract").

##### 1. Share of the borrowing interest (aka the "Reserve Factor").

When borrowers repay their loans, they do this with some interest.
Aave takes a portion of that interest to their treasury.

Technically, the interest amount ever grows, and Aave Treasury receives a portion of it:

https://github.com/aave/aave-v3-core/blob/782f51917056a53a2c228701058a6c3fb233684a/contracts/protocol/libraries/logic/ReserveLogic.sol#L264-L277

##### 2. Flash loan fee

Since the flash loans live only within a single transaction, a fee is applied instead of interest.
Aave takes a portion of this fee to their treasury.

https://github.com/aave/aave-v3-core/blob/782f51917056a53a2c228701058a6c3fb233684a/contracts/protocol/libraries/logic/FlashLoanLogic.sol#L230-L231

##### 3. Liquidation fee

Once borrower collateral is liquidated, the liquidator receives a share of the collateral as a premium.
Aave takes a portion of that premium to their treasury.

This fee is not shared back to depositors.

https://github.com/aave/aave-v3-core/blob/782f51917056a53a2c228701058a6c3fb233684a/contracts/protocol/libraries/logic/LiquidationLogic.sol#L514-L517

##### 4. Bridge fee

When the cross-chain bridge occurs, the fee is applied for issuing aTokens before the collateral is provided.
Aave takes a portion of that fee.

It _exists_, but I haven't found any application that uses it.

https://github.com/aave/aave-v3-core/blob/782f51917056a53a2c228701058a6c3fb233684a/contracts/protocol/libraries/logic/BridgeLogic.sol#L129-L140

#### Why estimation?

All the revenue generation described above could happen thanks to depositors, who provided liquidity to the protocol.
They receive their share expressed as ever-growing liquidity index.

Knowing the formulas above, we can try to answer the question: if the LP earned some amount, how much must the protocol earn? 

For simplicity's sake, I focused only on the borrowing interest, excluding all other fees from the scope.

If we consider only the borrowing interest, the calculation seems simple: if the protocol takes a share of the interest, and the rest goes to the depositors, we can express the depositor share as (1 - protocol share), and find the ratio of the protocol share to the depositor share: (protocol share / (1 - protocol share)). Then we could apply that ratio to the particular user earnings to _estimate_ the protocol share.

To estimate, because we ignore the flash loan fees and liquidation fees, essentially considering all protocol revenue coming from the interest.

I tend to think that these v0 calculations are a kind of proof of concept, demonstrating what is possible. And using this estimated metric as an example. We could enhance/change it in the future, should we see any interest from the community.

For instance, we may assume that the liquidity index, which basically reflects Aave's revenue sharing model, already includes all relevant components in the right proportions, and we could just multiply it to some rate we find appropriate.

#### Examples

I provided two examples of the calculation that could be replicated manually, in the test case:

https://github.com/divvi-xyz/divvi-protocol-v0/blob/1f4b9d5c9b493da0b6839f19820bf7ab56877418/scripts/calculateRevenue/protocols/aave/index.test.ts#L160-L183

#### See also

I asked ChatGPT for deep research on how Aave revenue is generated. It's an interesting high-level overview:

https://chatgpt.com/share/e/680b9985-bffc-8005-9566-a9f8db6c8f15

silasbw (approved):
we probably want some more careful audit of this before we ever use it.  Rather than leaving this PR open, let's merge it (and make sure it's at least up-to-date with refactorings, etc in this repo); or close it and re-consider it when we're ready to invest in a careful deepdive before using it.  I'll leave it up to you @bakoushin and @jeanregisser .

bakoushin:
I think we could treat this as an example of what is possible. And if we ever collaborate with Aave or related protocols, we can ask them for validation.

Another possible explanation for the proposed logic:
Aave already has a logic for sharing its revenue with the LPs.
Let's derive the referrer share from the LP share.
Let's use the Reserve Rate as a factor for that.

PR #94, opened 2025-04-11
feat: Add Fonbnk event filter

finnian0826:
Similar to the Aerodrome filtering, checks if the user has ever received a transfer from a Fonbnk payout wallet, and compares to the referral event timestamp.

Can break out of the pagination early because hypersync responses are in order, so will always find an invalid (too early) transaction first.


scripts/protocolFilters/fonbnk.ts
```diff
@@ -1,6 +1,82 @@
...
+    const payoutWallets = await Promise.all(
+      networkAssets.map((asset) =>
+        getPayoutWallets({ fonbnkNetwork: supportedNetwork, asset }),
+      ),
+    )
+    const uniquePayoutWallets = new Set(payoutWallets.flat())
+
+    // For each payout wallet, check that the user not transacted before the referral
+    // and for at least one of the payout wallets, the user has transacted
+    for (const payoutWallet of uniquePayoutWallets) {
```
kathaypacific, line 42: small nit: i might be missing some context here but it seems like you are getting all events from block 0 -> get the transaction -> for each transaction, get the block timestamp (async) -> check if there are events after the known timestamp. would it be more efficient to convert the known timestamp to a block number once per network, then make this query from that block? the check could then be simplified to return true if there any events found? converting a timestamp to a block number isn't very trivial but we have this `getNearestBlock` utility function which may help...

scripts/protocolFilters/fonbnk.ts
```diff
@@ -1,6 +1,82 @@
...
+  if (!isAddress(event.userAddress)) {
+    throw new Error(`Invalid user address: ${event.userAddress}`)
+  }
+
+  let foundValidTransaction = false
+  let transactedBeforeReferral = false
+
+  const fonbnkAssets = await getFonbnkAssets()
+  // For each network, create a new hypersync client and get all of the unique payout wallets
+  for (const supportedNetwork of SUPPORTED_FONBNK_NETWORKS) {
```
kathaypacific, line 26: there are pros and cons to serialising the check per network, i'm not sure which is more optimal for fonbnk but wanted to call out that if it may be faster overall if we parallelised the work for all networks with Promise.all, and do a final check that _any_ result is true. there are at least 2 async tasks per network (to get the payout wallets, and to query the indexer) so it may be a noticeable performance lift? i'm not sure

scripts/protocolFilters/fonbnk.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      (_networkId: NetworkId, _blockNumber: bigint) =>
+        Promise.resolve({
+          timestamp: 1742194800n, // March 17, 2025
+        }) as unknown as ReturnType<typeof getBlock>,
+    )
+
+    const result = await filter(event)
+    expect(result).toBe(false)
+    expect(mockClient.get).toHaveBeenCalledTimes(1)
+    expect(getBlock).toHaveBeenCalled()
```
kathaypacific, line 69: nit: if we want the tests to provide good coverage for how `getBlock` is called, we can either do `expect(getBlock).toHaveBeenCalledWith(...)` or in the mock implementation, throw an error if the networkId / blockNumber are not what we expect. with the current tests, i could accidentally leave a hard coded value in the input arg of `getBlock` (from debugging, idk) and the tests would not catch that

scripts/protocolFilters/fonbnk.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    const result = await filter(event)
+    expect(result).toBe(true)
+    expect(mockClient.get).toHaveBeenCalledTimes(1)
+    expect(getBlock).toHaveBeenCalled()
+  })
+
+  it('returns false if no blocks are found', async () => {
+    mockClient.get
+      .mockResolvedValueOnce(makeQueryResponse([]))
+      .mockResolvedValueOnce(makeQueryResponse([])) // causes loop to break
```
kathaypacific, line 93: what does this comment mean? should mockClient have been called 2 times or is this some mocking defect?

kathaypacific (approved):
LGTM!

scripts/protocolFilters/fonbnk.test.ts
finnian0826: This is a result of how the pagination added by Tom works, I believe with mocking it like this, it will get the first value, then try to get another one and see that `nextBlock` is the same as in the previous response so then will stop pagination. These tests are mostly copied over and modified from the ones Tom added for Aerodrome.

scripts/protocolFilters/fonbnk.ts
kathaypacific: OOHHHHH i misunderstood, what you're trying to do is find the _first_ transaction the user made with fonbnk, and verify if it is after the referral was made? in that case, do you need both `transactedBeforeReferral` and `foundValidTransaction`? perhaps you could check only the first transaction returned by the query then to simplify the logic?

scripts/protocolFilters/fonbnk.ts
finnian0826: created a follow-up ticket to optimize more [here](https://linear.app/valora/issue/ENG-295/placeholder-optimize-fonbnk-revenue-calculation-and-filtering), for both the revenue calculation and filtering if they end up being too slow. Hopefully current implementation is okay for the async calculation done in v0 though.

PR #98, opened 2025-04-14
chore(aave): data fetching

bakoushin:
## Description

This PR is focused on blockchain data fetching required for the Aave revenue calculation.

The data includes:

### 1. Reserve tokens

* list of all reserve tokens supported by a specific Aave pool, to check user balances against
* their properties like decimals
* their corresponding ATokens (see below)

### 2. AToken scaled balance history
* starting block scaled balance
* scaled balance changes during the selected period

AToken is given to depositors in exchange for their deposits. When withdrawing, they exchange their ATokens for the deposited funds. The exchange rate is always 1:1.

Internally, Aave tracks the _scaled balance_ of the ATokens. When user deposit, their scaled balance is calculated as the amount being deposited divided by the liquidity index at this moment. When a user withdraws, their scaled balance is multiplied by the liquidity index at the moment of withdrawal, yielding them more ATokens than at the initial deposits. The liquidity index is constantly growing, representing the growth of depositor earnings.

> **Liquidity Index** Tracks the cumulative interest earned by a reserve over time, used to calculate accurate interest payments.

### 3. Reserve factor history

* starting block reserve factor
* reserve factor changes during the selected period

Reserve factor is the protocol share in earnings.

> **Reserve Factor** A percentage of interest accrued by borrowers that is allocated to the Aave Treasury to help safeguard the protocol. Current reserve factors: https://aave.com/docs/resources/parameters

### 4. Reserve token USD prices

* USD prices stored in Aave oracle at the end block

We will use these prices to convert the protocol revenue represented in each reserve token to USD to be able to calculate the total amount of protocol revenue across all reserves.

## Sources of data

* The starting/ending values we get from the RPC call at the starting/ending block
* The reserve factor history we get from Hypersync query tracking the `ReserveFactorChanged` event
* The AToken balance history we get from an Aave subgraph, as it is way simpler than tracking individual raw Deposit/Withdraw/Transfer events

## Related issues

Related to ENG-251


scripts/calculateRevenue/protocols/aave/blockchainData/oracle.ts
```diff
@@ -0,0 +1,64 @@
...
+  const baseCurrencyUnit = poolByteCode
+    ? await publicClient.readContract({
+        address: oracleAddress,
+        abi: oracleAbi,
+        functionName: 'BASE_CURRENCY_UNIT',
+        blockNumber: BigInt(blockNumber),
+      })
+    : 0n
+
+  const prices = poolByteCode
```
kathaypacific, line 41: i wondered if we should parallelise these contract reads, then i remembered we have the batching set on the viem level! cool

kathaypacific (approved):
🚀

scripts/calculateRevenue/protocols/aave/types.ts
```diff
@@ -0,0 +1,20 @@
...
+  timestamp: number
+}
+
+export interface ReserveData {
+  reserveTokenAddress: Address
+  reserveTokenDecimals: number
+  aTokenAddress: Address
+  liquidityIndex: bigint
+  reserveFactor: bigint
+}
```
jeanregisser, line 20: It would be helpful to explain the fields, possibly linking to aave docs.
So next time someone needs to look at this their work will be facilitated.

scripts/calculateRevenue/protocols/aave/blockchainData/subgraph.ts
```diff
@@ -0,0 +1,89 @@
...
+      userReserves(where: { user: $userAddress }) {
+        reserve {
+          aToken {
+            id
+          }
+        }
+        aTokenBalanceHistory(
+          where: {
+            timestamp_gte: $startTimestamp
+            timestamp_lte: $endTimestamp
```
jeanregisser, line 58: Depending on whether we actually follow the inclusive-start, exclusive-end convention for ranges, this should be `timestamp_lt`.

PR #99, opened 2025-04-14
feat: Add velodrome referral event filtering

finnian0826:
Added a generic `filterDrome` function, then used for both aero and velodrome.


scripts/utils/filterDrome.ts
```diff
@@ -0,0 +1,37 @@
+import { BlockField } from '@envio-dev/hypersync-client'
+import { NetworkId, ReferralEvent } from '../types'
+import { getBlock, getHyperSyncClient } from '../utils'
+import { paginateQuery } from '../utils/hypersyncPagination'
+
+export async function filterDrome({
+  event,
+  routerAddress,
+  networkId,
+}: {
```
finnian0826, line 10: This is just copied over from the aerodrome filtering, and props added to distinguish between aerodrome or velodrome

MuckT (approved):

PR #100, opened 2025-04-16
feat: add referral management to DivviRegistry

kathaypacific:
The scope of this PR:
- setting, getting, storing referrals
- method for batching referrals (not included in the task but I think we need this?)
- tests

Not much to call out, I followed the recommendation pretty closely aside from modifying some variable names.

Related to ENG-298


contracts/DivviRegistry.sol
```diff
@@ -31,6 +39,21 @@ contract DivviRegistry is
...
+    bytes32 txHash
+  );
+  event ReferralSkipped(
+    address indexed user,
+    address rewardsProvider,
+    address indexed rewardsConsumer,
+    uint256 chainId,
+    bytes32 indexed txHash,
+    string reason
+  );
```
jeanregisser, line 56: Regarding the `ReferralSkipped` event: emitting `string reason` is readable but more gas-intensive, especially within `batchRegisterReferral`.

I recommend replacing it with the `ReferralStatus status` enum value for better gas efficiency. Off-chain systems can still map the status code to a description.

Let me know your thoughts.

contracts/DivviRegistry.sol
```diff
@@ -31,6 +39,21 @@ contract DivviRegistry is
     address indexed rewardsProvider,
     address indexed rewardsConsumer
   );
+  event ReferralRegistered(
+    address user,
+    address indexed rewardsProvider,
+    address indexed rewardsConsumer,
+    uint256 indexed chainId,
```
jeanregisser, line 46: I'm still wondering whether we should use our string network ID (`ethereum-mainnet`, `op-mainnet`, etc). 🤔 

It is more gas efficient to use the numeric version.
But wondering if we want to support non EVM chains.

Well, I see that protocols like hyperlane which support Solana [assigned it a chainId too](https://github.com/hyperlane-xyz/hyperlane-monorepo/blob/0e663ad85ddb681b684d80212039e8ae5c82b1c7/rust/main/hyperlane-core/src/chain.rs#L171C21-L171C31).
And I think Wormhole and LayerZero do something similar.

jeanregisser (commented):
Awesome! 🙌

contracts/DivviRegistry.sol
kathaypacific: yes it's the readability that made me lean towards a string, but you're right that anyone who is interested can map the enum to a string reason if they need it. gas efficiency isn't a good tradeoff then, i updated to use the enum

contracts/DivviRegistry.sol
```diff
@@ -37,6 +60,21 @@ contract DivviRegistry is
...
+    bytes32 txHash
+  );
+  event ReferralSkipped(
```
kathaypacific, line 70: my understanding of this "indexed" keyword is that it makes it faster to query the blockchain for events with this property, but i'm not really sure how we might want to analyse these events. there can only be maximum 3 indexed properties per event, so i've chosen to be able to quickly see how often a consumer is unsuccessful in the referral / how often a user and tx hash is bouncing. this info would help to raise awareness of bugs maybe. open for alternative preferences for which properties are indexed

jophish: Hmm I would probably advocate for indexing `user` `rewardsProvider` and `rewardsConsumer`, since this lets us segment events across each of the main parties we care about in the protocol. e.g., we can easily see:
* All the entities a user is referred to
* All the users referred to a particular entity
* All the users referred by a particular entity

If we want to look up by `txHash`, and we know the user who submitted the `txHash` (which we always should, since we can just check the sender of the transaction), we can just look up events by user (should be a relatively small number per user) and filter them down.

contracts/DivviRegistry.sol
```diff
@@ -159,6 +197,91 @@ contract DivviRegistry is
...
+   * @notice Register a user as being referred to a rewards agreement
+   * @dev Internal function that returns status instead of emitting events
+   * @param user The address of the user being referred
+   * @param rewardsProvider The address of the rewards provider entity
+   * @param rewardsConsumer The address of the rewards consumer entity
+   * @param txHash The hash of the transaction that initiated the referral
+   * @param chainId The ID of the blockchain where the referral transaction occurred
+   * @return status The status of the referral registration
+   */
+  function _registerReferral(
```
jophish, line 252: Hmm, I notice that `txHash` and `chainId` are totally unused throughout the contract besides being emitted in events... is it necessary to provide them to the contract at all, since our registrar address is the only address permitted to submit referrals, and we'll be verifying transactions ourselves before submitting referrals to the contract?

contracts/DivviRegistry.sol
```diff
@@ -21,6 +21,14 @@ contract DivviRegistry is
   mapping(bytes32 => bool) private _agreements; // keccak256(provider, consumer) => true (if agreement exists)
   mapping(address => bool) private _requiresApproval; // entity => boolean (if entity requires approval)
 
+  // Referral tracking
+  mapping(bytes32 => address) private _registeredReferrals; // keccak256(user, provider) => consumer
+  mapping(bytes32 => bool) private _registrationTransactions; // keccak256(chainId, txHash) => true (if the transaction has been used for registering a referral)
```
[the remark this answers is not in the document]
jeanregisser: Yes this is perfect!

contracts/DivviRegistry.sol
jeanregisser: Yes I think this is good to have as an audit trace.
"ok the referral verifier is centralized for now, but anyone can externally see it correctly attributes referrals, by having a link to the original TX containing the referral data tag".

jeanregisser (approved):
+1 to what Joe said.

Already approved for the rest.

contracts/DivviRegistry.sol
kathaypacific: that makes sense, thanks for the reasoning! updated!

kathaypacific: i also updated ReferralRegistered to have the same for consistency, i hope this is expected

PR #104, opened 2025-04-23
chore: add deploy tasks for DivviRegistry

kathaypacific:
This PR:
- adds deploy and upgrade tasks for DivviRegistry (moving some checks around to reduce duplicated code)
- updates README
- removes unused arguments in DivviRegistry

Related to ENG-234

### DIFF
--- a/tasks/helpers/deployHelpers.ts
+++ b/tasks/helpers/deployHelpers.ts
@@ -20,9 +20,16 @@ type BaseDeployConfig = {
   isUpgradeable?: boolean
 }
 
-type DefenderConfig = {
-  useDefender?: boolean
+type DefenderConfig = WithDefender | WithoutDefender
+
+type WithDefender = {
+  useDefender: true
   defenderDeploySalt?: string
+}
+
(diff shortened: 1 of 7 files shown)


tasks/helpers/deployHelpers.ts
```diff
@@ -20,24 +20,30 @@ type BaseDeployConfig = {
   isUpgradeable?: boolean
 }
 
-type DefenderConfig = WithDefender | WithoutDefender
```
kathaypacific, line 23: without this change, typescript complains about the check for `config.defenderDeploySalt && !config.useDefender` - we could repeat the checks in every deploy task or loosen the type definition here, i oped for the latter to reduce duplicated code

sviderock: @kathaypacific shouldn't it work if you change `WithoutDefender` to this?
```ts
type WithoutDefender = {
  useDefender?: false
  defenderDeploySalt?: never
}
```

jeanregisser (approved):
🚀

PR #109, opened 2025-04-24
feat: add gasless tx support via ERC2771

jeanregisser:
## Description

Adds support for **gasless transactions** to the `DivviRegistry` using the standard [ERC-2771](https://eips.ethereum.org/EIPS/eip-2771) meta-transaction approach.

> [!TIP]
> Please review using the "Hide whitespace" diff option to more easily see the important changes.

Fixes ENG-309

## Why is this important?

Right now, builders need native tokens (like ETH) to pay gas for any interaction with the registry (registering, creating agreements, etc.). This PR makes it possible for us (or other relayers we trust) to **sponsor those transactions**. Builders can then interact with the registry without worrying about gas, which is a much better UX.

Basically, any interaction with the registry can now potentially be gas-free for the end-user. Since we're using the standard ERC-2771 pattern, this should play nicely with common relay services like Gelato Relay right out of the box (using their `sponsoredCallERC2771` function, for example).

*   Gelato Docs: https://docs.gelato.network/web3-services/relay/quick-start/sponsored-calls

## How's it implemented?

Using OpenZeppelin's helpers:

*   Pulled in `ERC2771ContextUpgradeable` to handle the core logic. (OZ Docs: [https://docs.openzeppelin.com/contracts/5.x/api/metatx](https://docs.openzeppelin.com/contracts/5.x/api/metatx))
*   Swapped out `msg.sender` for `_msgSender()` everywhere. This function figures out the *real* sender when a call comes through a trusted forwarder.
*   Added a `TRUSTED_FORWARDER_ROLE`. Instead of hardcoding a forwarder address, `isTrustedForwarder()` now just checks if the calling address has this role. This lets us easily manage which relayers we trust via the normal admin roles.

## Security notes

*   **Trust is key:** The whole system relies on trusting the forwarder contract. We manage this trust using the `TRUSTED_FORWARDER_ROLE`. Only admins can grant this role.
*   **Immutable forwarders only:** We should *only* ever grant the `TRUSTED_FORWARDER_ROLE` to forwarder contracts that are **immutable** and preferably audited. If the forwarder could be upgraded or was dodgy, it could cause all sorts of problems.
*   **Gelato example:** Gelato's forwarders are a good example, they are typically immutable contracts deployed specifically for this. Like this one on Optimism: [0xc65d82ECE367EF06bf2AB791B3f3CF037Dc0e816](https://optimistic.etherscan.io/address/0xc65d82ECE367EF06bf2AB791B3f3CF037Dc0e816#code). See all their supported deployments: https://docs.gelato.network/web3-services/relay/supported-networks.

## Testing

*   Updated the tests structure. Introduced a helper (`executeAs`) to easily run test cases both as direct calls *and* as meta-transactions coming from our simulated trusted forwarder.
*   All the main functions (`registerRewardsEntity`, `setRequiresApproval...`, `registerAgreement...`, `batchRegisterReferral`, even `grantRole`) are now covered in both modes.
*   Added a new test specifically to make sure things revert correctly if someone tries to relay a meta-tx through an *untrusted* address.


kathaypacific (approved):
🚀 wow neat! i also like the `executeAs` function in the tests

PR #111, opened 2025-04-28
feat(contracts): Implement initial Data Availability contract

jophish:
For context on the full flow, see [this writeup](https://github.com/divvi-xyz/risc0-example/pull/3/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R6).

Adds the initial version of the Data Availability contract. As a best-practice, this contract does not actually store the data, but instead emits events for each "upload" such that the history can be fully re-created via indexing. Instead of storing the data, the contract maintains a streaming hash of the data at each timestamp by XORing the hash of each user:value pair. Since XOR is commutative, this allows an uploader to upload data for a particular timestamp in any order, and the final hash will always be the same.

The verification process will look as follows:
* In a trusted zkVM, calculations will be performed on raw data in order to generate the list of user:value pairs stored in the contract.
* The zkVM will compute the hash of these user:value pairs in the same way the contract does.
* The zkVM generates a proof of this computation, and attaches the hash to the journal.
* This proof is uploaded to the contract along with the hash from the trusted compute environment
* The contract verifies the proof, and compares the journal hash to the contract hash for the given timestamp.
* If they match, we can be confident that the trusted code running in the zkVM produced the same data as those previously streamed to the contract, and can mark the contract data as verified.


contracts/DataAvailability.sol
```diff
@@ -0,0 +1,212 @@
...
+     * but events are emitted for each upload, which can be used to reconstruct the full data history.
+     *
+     * Instead of storing all data in the contract, a rolling hash of the data is calculated and stored
+     * for each timestamp. When verifying the data, the submitted proof will contain a hash of the data
+     * computed in a trusted zkVM; if the hash contained in the proof matches the hash stored in the contract,
+     * the data is considered valid.
+
+     * The rolling hash is calculated by XORing the existing hash for a given timestamp with the hash of each
+     * user:value pair as they are received by the contract. Since XOR is both commutative and associative, this
+     * results in a hash that is invariant to the order of uploads. Since we cannot "remove" information about
```
jophish, line 24: This is technically not true; we can "remove" a piece of data from the hash's history by just uploading it again, since A ^ A = 0, which would allow us to "correct" it later. This seems like a weird thing to do though, so I just decided to prevent that case here.

test/DataAvailability.ts
```diff
@@ -0,0 +1,281 @@
...
+      const values = [200, 100]
+      await dataAvailability
+        .connect(uploader)
+        .uploadData(timestamp, users, values)
+
+      const expectedHash =
+        '0x21222613b8f454858d6b5c0b2be37eb83f88affe0bef993fabf01e4e5faf8ad8'
+      const hash = await dataAvailability.getHash(timestamp)
+      expect(hash).to.equal(expectedHash)
+    })
```
jh2oman, line 219: Test would probably be better if you just calculated each hash and asserted that they were equal instead of matching the `expectedHash`

jophish: yeah fair enough, i'll update to actually perform the calc within the test itself

contracts/DataAvailability.sol
```diff
@@ -0,0 +1,208 @@
...
+     * The contract creator and any permitted uploaders may upload user objective function data
+     * at a given timestamp. The objective function data that are uploaded represent the delta
+     * between the objective function value at the current timestamp and the value at the
+     * previous timestamp that data was uploaded. The data is not actually stored in the contract,
+     * but events are emitted for each upload, which can be used to reconstruct the full data history.
+     *
+     * Instead of storing all data in the contract, a rolling hash of the data is calculated and stored
+     * for each timestamp. When verifying the data, the submitted proof will contain a hash of the data
+     * computed in a trusted zkVM; if the hash contained in the proof matches the hash stored in the contract,
+     * the data is considered valid.
```
jh2oman, line 20: What would happen if the hash stored in the contract and the hash contained in the proof don't match and the data is now considered invalid?

contracts/DataAvailability.sol
```diff
@@ -0,0 +1,225 @@
...
+  ) external onlyRole(UPLOADER_ROLE) {
+    require(users.length == values.length, 'Arrays length mismatch');
+
+    // Ensure the timestamp is greater than the most recent timestamp
+    if (timestamps.length > 0) {
+      require(
+        timestamp >= timestamps[timestamps.length - 1],
+        'Timestamp must be greater than or equal to most recent timestamp'
+      );
+    }
```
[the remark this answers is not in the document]
jophish: There are kind of weird consequences if we allow "retroactive" updates for older timestamps. Imagine we have t0 < t1< t2 < t3 , and the following happens:

* Entity uploads data for t1 and verifies it
* Our rewards script looks at t1 data and sends rewards (These are the rewards from t0 to t1)
* Entity uploads data for t3 and verifies it 
* Our reward script looks at t3 data and sends rewards (These are the rewards from t1 to t3)

Now, if data is uploaded for t2, things get kinda weird -- our rewards script won't be able to pick this up by looking at the latest timestamp that data is available for. More importantly, the data originally uploaded for t3 represents the increase in the KPI since t1; uploading data for t2 changes the semantics of this t3 data to now represent the increase in the KPI since t2, which would make the t3 data now incorrect given the new semantics.

contracts/DataAvailability.sol
```diff
@@ -0,0 +1,225 @@
...
+
+  /**
+   * @dev Get the stored hash for a given timestamp
+   * @param timestamp The timestamp to query
+   * @return The stored hash for the given timestamp
+   */
+  function getHash(uint256 timestamp) external view returns (bytes32) {
+    if (!this.hasDataForTimestamp(timestamp)) {
+      return bytes32(0);
+    }
```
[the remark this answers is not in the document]
jophish: Ah yeah, this is superfluous, good call!

contracts/DataAvailability.sol
```diff
@@ -0,0 +1,225 @@
...
+     * between the objective function value at the current timestamp and the value at the
+     * previous timestamp that data was uploaded. The data is not actually stored in the contract,
+     * but events are emitted for each upload, which can be used to reconstruct the full data history.
+     *
+     * Instead of storing all data in the contract, a rolling hash of the data is calculated and stored
+     * for each timestamp. When verifying the data, the submitted proof will contain a hash of the data
+     * computed in a trusted zkVM; if the hash contained in the proof matches the hash stored in the contract,
+     * the data is considered valid.
+
+     * The rolling hash is calculated by XORing the existing hash for a given timestamp with the hash of each
```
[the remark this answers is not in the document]
jophish: Ohh interesting, do you mean something like:

```
hash = (keccak256(chunk1) + hash) % 2^256
```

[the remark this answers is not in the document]
jophish: This apparently has better collision resistance than the simple XOR, so I'm opting for this instead.

contracts/DataAvailability.sol
```diff
@@ -0,0 +1,223 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import {AccessControlDefaultAdminRules} from '@openzeppelin/contracts/access/extensions/AccessControlDefaultAdminRules.sol';
+
+contract DataAvailability is AccessControlDefaultAdminRules {
+  /**
+     * @dev DataAvailability stores information about the objective function value for users
```
silasbw, line 8: Can you give context/example on how a reward producer might use this?  E.g., "deploy a new instance for each 'KPI'"

contracts/DataAvailability.sol
```diff
@@ -0,0 +1,223 @@
...
+  constructor(address _owner) AccessControlDefaultAdminRules(0, _owner) {
+    // Grant the deployer the uploader role
+    _grantRole(UPLOADER_ROLE, _owner);
+  }
+
+  /**
+   * @dev Override to prevent the owner from losing their uploader role
+   */
+  function revokeRole(bytes32 role, address account) public virtual override {
+    require(
```
silasbw, line 65: we usually do `revert CustomError`.  We had that enabled as a linter option at one point too.  Should we follow the "revert pattern"?

silasbw (commented):
nits

silasbw (approved):

PR #113, opened 2025-04-29
feat: add total celo transaction calculator

satish-ravi:


scripts/types.ts
```diff
@@ -7,6 +7,7 @@ export const protocols = [
   'velodrome',
   'fonbnk',
   'aave',
+  'celo-transactions',
```
satish-ravi, line 10: open to naming suggestions

silasbw: sounds good -- could we also make, like, an alias of `celo` called `celo-transactions` for calling fetchReferrals?

silasbw (approved):

scripts/types.ts
satish-ravi: yes good idea, I can make this with the fetch referrals update

PR #117, opened 2025-05-02
feat: Add script to get entities that have integrated with Divvi

finnian0826:
We want to reward entities who have successfully registered a user with Divvi. We only want to reward them once.

This script will find the addresses of all entities that have referred a user, and have not yet received a reward. It will output these addresses into the output csv.

```
yarn ts-node ./scripts/getDivviIntegrators.ts --reward-pool=${RewardPool contract address} --reward-amount=${amount in USD}
```
will output a csv of `address, amount` in `divvi-integrator-rewards.csv`


scripts/getDivviIntegrators.ts
```diff
@@ -0,0 +1,168 @@
...
+    paginateQuery(client, queryForRegisteredAgreements, async (response) => {
+      for (const transaction of response.data.logs) {
+        usersThatHaveRegisteredAgreements.add(transaction.topics[2] as Address)
+      }
+    }),
+  ])
+
+  const deduplicatedUsersThatHaveIntegrated = removeDuplicates(
+    usersThatHaveIntegrated,
+  )
```
finnian0826, line 137: users may have referred more than one user so could show up more than once, remove duplicates while retaining order since we want to reward the first X builders

scripts/getDivviIntegrators.ts
```diff
@@ -0,0 +1,168 @@
...
+}
+
+async function getDivviIntegrators({
+  rewardPoolContractAddress,
+}: {
+  rewardPoolContractAddress: string
+}): Promise<Address[]> {
+  const usersThatHaveIntegrated: Address[] = []
+  const usersThatHaveReceivedRewards = new Set<Address>()
+  const usersThatHaveRegisteredAgreements = new Set<Address>()
```
kathaypacific, line 59: can we change `user` -> `consumer` to be consistent with the terminology (since `user` should refer to the end user, this was a bit confusing to read)

scripts/getDivviIntegrators.ts
```diff
@@ -0,0 +1,168 @@
...
+  const deduplicatedUsersThatHaveIntegrated = removeDuplicates(
+    usersThatHaveIntegrated,
+  )
+
+  // TODO(ENG-345): Also filter for if the user is whitelisted
+  const userToReceiveRewards = deduplicatedUsersThatHaveIntegrated.filter(
+    (user: Address) =>
+      !usersThatHaveReceivedRewards.has(user) &&
+      usersThatHaveRegisteredAgreements.has(user),
+  )
```
kathaypacific, line 144: do we need to track the registered agreements? the contract already ensures that [there is an agreement](https://github.com/divvi-xyz/divvi-protocol-v0/blob/57649c63778eccbfa05c0b4dcfd798e89ae5ba24/contracts/DivviRegistry.sol#L268-L273) before registering the referral

scripts/getDivviIntegrators.ts
```diff
@@ -0,0 +1,168 @@
...
+    })
+    .option('reward-pool', {
+      description: 'RewardPool contract address',
+      type: 'string',
+      demandOption: true,
+    })
+    .option('reward-amount', {
+      description: 'reward amount for integration',
+      type: 'string',
+    }).argv
```
kathaypacific, line 32: should we allow passing in the start and end blocks, to avoid doing extra work between script executions?

kathaypacific (commented):
🙏🏻

scripts/getDivviIntegrators.ts
kathaypacific: ooohhhh i see, `usersThatHaveRegisteredAgreements` is to track which entities have agreements specifically with the divvi provider. can we rename the variable to be more specific? e.g. `divviAgreementConsumers` / `consumersWithDivviAgreements`?

PR #121, opened 2025-05-05
chore: Add allowlist check to divvi integrators reward script

finnian0826:
Fetch the allowlist from the integration-list repo, and check that the builder is in it.


scripts/getDivviIntegrators.ts
```diff
@@ -7,12 +7,20 @@ import { getHyperSyncClient } from './utils'
...
 import { divviRegistryAbi } from '../abis/DivviRegistry'
 import { rewardPoolAbi } from '../abis/RewardPool'
+import { fetchWithTimeout } from './utils/fetchWithTimeout'
 
 const DIVVI_REGISTRY_CONTRACT_ADDRESS =
   '0xEdb51A8C390fC84B1c2a40e0AE9C9882Fa7b7277'
 const DIVVI_INTEGRATION_REWARDS_ENTITY =
   '0x6226ddE08402642964f9A6de844ea3116F0dFc7e'
 
+const WHITELIST_URL =
```
jeanregisser, line 17: nit: consider replacing `whitelist` with `allowlist` to align with inclusive language best practices. It’s clearer and more widely adopted across the industry now.

jeanregisser (approved):
🚀

scripts/getDivviIntegrators.ts
```diff
@@ -136,11 +150,25 @@ async function getDivviIntegrators({
     usersThatHaveIntegrated,
   )
 
-  // TODO(ENG-345): Also filter for if the user is whitelisted
+  const fetchAllowlistResponse = await fetchWithTimeout(ALLOWLIST_URL)
```
kathaypacific, line 153: could we create a function that fetches this allow list (including the error handling / data massaging) and put this async function in the Promise.all above? i guess this api request will be fast, but just in case it's not?

kathaypacific (approved):

PR #131, opened 2025-05-09
feat: add safe transactions batch to calculate rewards

kathaypacific:
This PR repurposes the batch safe transaction script from the v0 registrations to be used for generating the transactions for adding rewards. The intention is that reward calculation functions will call this utility method to generate a JSON file that we can upload to the Safe transaction builder to add the rewards.

https://docs.google.com/document/d/1rKIKMtlz3TKFkHXmNqD4B7VH10GB4MUUhl1MKzdnu88/edit?tab=t.0

Related to ENG-332


jophish (approved):
Looks great!!! Very satisfying to see all the pieces coming together :)

PR #137, opened 2025-05-13
fix: calculateRevenue should look at referral timestamp

kathaypacific:
https://docs.google.com/document/d/1rKIKMtlz3TKFkHXmNqD4B7VH10GB4MUUhl1MKzdnu88/edit?disco=AAABg72mb6E


scripts/calculateRevenue.ts
```diff
@@ -23,13 +26,21 @@ async function main(args: ReturnType<typeof parseArgs>) {
   }> = []
 
   for (let i = 0; i < eligibleUsers.length; i++) {
-    const { referrerId, userAddress } = eligibleUsers[i]
+    const { referrerId, userAddress, timestamp } = eligibleUsers[i]
     console.log(
       `Calculating revenue for ${userAddress} (${i + 1}/${eligibleUsers.length})`,
     )
+
+    const referralDate = new Date(timestamp - REFERRAL_TIME_BUFFER_IN_SECONDS)
```
jophish, line 34: This might be overkill/unnecessary, but is it worth skipping users whose `referralDate > endTimestamp`? It's possible(?) that if a user like this sneaks in, things will break/throw within the revenue calculation since `startTimestamp > endTimestamp` in that case.

jophish (approved):

scripts/calculateRevenue.ts
```diff
@@ -23,13 +26,21 @@ async function main(args: ReturnType<typeof parseArgs>) {
...
+
     const revenue = await handler({
       address: userAddress,
-      startTimestamp: new Date(args['start-timestamp']),
+      // if the referral happened after the start of the period, only calculate revenue from the referral block onwards so that we exclude user activity before the referral
+      startTimestamp:
+        referralDate.getTime() > periodStartDate.getTime()
+          ? referralDate
+          : periodStartDate,
       endTimestamp: new Date(args['end-timestamp']),
```
jeanregisser, line 44: The updated code I'm working ensure startTimestamp < endTimestamp and throws if that's not the case.
What do you think of skipping early if `referralDate.getTime() >= endTimestamp`? Possibly logging something if useful.
And avoid the future exception.

jeanregisser (approved):
🚀

scripts/calculateRevenue.ts
kathaypacific: yes it's possible since we'll run the calculations for a past time period, i overlooked this scenario! thanks, updated :D

PR #139, opened 2025-05-14
chore: hard code some getDivviIntegrators.ts parameters and add safe txs

silasbw:
We can add the parameters back if we want to include more support, but it felt like too many sharp edges when writing up the process for distributing.


scripts/getDivviIntegrators.ts
```diff
@@ -9,12 +16,19 @@ import { divviRegistryAbi } from '../abis/DivviRegistry'
...
-
+import { createAddRewardSafeTransactionJSON } from './utils/createSafeTransactionsBatch'
 const DIVVI_REGISTRY_CONTRACT_ADDRESS =
   '0xEdb51A8C390fC84B1c2a40e0AE9C9882Fa7b7277'
 const DIVVI_INTEGRATION_REWARDS_ENTITY =
   '0x6226ddE08402642964f9A6de844ea3116F0dFc7e'
 
+// TODO(sbw): hardcoded for now, but if we launch more campaings we can address TODOs
+// to take this as a CLI argument.
+const DIVVI_REWARD_POOL_ADDRESS = '0x326161d68c05bE55367a0041b9C8f68082C04863'
```
satish-ravi, line 27: I guess hardcoding is explicit and less error-prone, but another option is to leave the arg option and set defaults

satish-ravi (approved):

scripts/getDivviIntegrators.ts
kathaypacific: i have a weak preference for not having optional args because it requires you to know what the default is in order to know whether it's safe to not provide it, which is extra mental overhead + increases the possibility of accidentally not providing the arg and using the default incorrectly?

kathaypacific (approved):

scripts/getDivviIntegrators.ts
silasbw: ...If we did the default suggestion, we'd also need to handle the decimals do we didn't specify the wrong decimals for the rewardpool/reward token.

PR #140, opened 2025-05-14
feat: first reward distribution divvi integrator

kathaypacific:
Context: https://valora-app.slack.com/archives/C02898GN22V/p1747198071228809?thread_ts=1747194569.415919&cid=C02898GN22V

This PR adds the reward transaction that will be uploaded to the divvi integrator reward pool. To verify the correctness of the transaction data:
1. check out the main branch locally, ensure you are up to date
2. run `yarn ts-node ./scripts/getDivviIntegrators.ts --start-timestamp=1745107200000 --end-timestamp=1747181107000 --safe-transactions-file=rewards/divvi-integrators/1745107200000-1747181107000/safe-transactions.json`
3. verify that the generated json matches this PR


scripts/utils/createSafeTransactionsBatch.ts
```diff
@@ -61,7 +62,9 @@ export const createAddRewardSafeTransactionJSON = ({
     ],
   }
 
-  writeFileSync(filePath, JSON.stringify(transactionsBatch, null, 2), {
+  // Create directory if it doesn't exist
+  mkdirSync(dirname(filePath), { recursive: true })
+  writeFileSync(filePath, JSON.stringify(transactionsBatch, null, 2) + '\n', {
```
kathaypacific, line 67: add extra new line so that prettier doesn't complain

jeanregisser (approved):
Verified locally that it matches.

One small question, should we also commit the `divvi-integrator-rewards.csv` to the rewards folder?
```
$ /Users/jean/src/github.com/divvi-xyz/divvi-protocol-v0/node_modules/.bin/ts-node ./scripts/getDivviIntegrators.ts --start-timestamp=1745107200000 --end-timestamp=1747181107000 --safe-transactions-file=rewards/divvi-integrators/1745107200000-1747181107000/safe-transactions.json
Querying for consumers...
Found 5 consumers that have integrated
Found 0 consumers that have received rewards
Found 35 consumers with Divvi integration agreement
Found 2 consumers that are allowlisted
Wrote results to divvi-integrator-rewards.csv
```

kathaypacific:
> Verified locally that it matches.
> 
> One small question, should we also commit the `divvi-integrator-rewards.csv` to the rewards folder?
> 
> ```
> $ /Users/jean/src/github.com/divvi-xyz/divvi-protocol-v0/node_modules/.bin/ts-node ./scripts/getDivviIntegrators.ts --start-timestamp=1745107200000 --end-timestamp=1747181107000 --safe-transactions-file=rewards/divvi-integrators/1745107200000-1747181107000/safe-transactions.json
> Querying for consumers...
> Found 5 consumers that have integrated
> Found 0 consumers that have received rewards
> Found 35 consumers with Divvi integration agreement
> Found 2 consumers that are allowlisted
> Wrote results to divvi-integrator-rewards.csv
> ```

i wondered about this too! there is a .gitignore line that ignores all .csv files though. 

i was thinking that the only reason we'd want to keep metadata is for easy reviews / storing data that is a function of time (e.g. api responses). i was thinking for this particular calculation that the consumers are fetched from the blockchain and the allowlist is fetched from the main branch of the divvi-integrators GH project, and given the start / end time we should be able to fetch the same results every time. so i landed on it not being necessary to store the intermediate .csv file....

looking through the script now i wonder if the .csv file was a temporary "finish line" for the script, before we decided on the safe transaction json thing...

rewards/divvi-integrators/1745107200000-1747181107000/safe-transactions.json
```diff
@@ -0,0 +1,36 @@
+{
+  "meta": {},
+  "transactions": [
+    {
+      "to": "0x326161d68c05bE55367a0041b9C8f68082C04863",
```
jeanregisser, line 5: Is this the right address?
CeloScan says it's an EOA, is it an indexing issue?
https://celoscan.io/address/0x326161d68c05bE55367a0041b9C8f68082C04863

kathaypacific: ah yes it's optimism 🙈

PR #141, opened 2025-05-14
fix(scripts): introduce getBlockRange and standardize to [start,end) block ranges

jeanregisser:
As the title says. 

This fixes subtle off-by-one errors we had due to HyperSync / RPC node not treating the "to block" the same way (exclusive vs inclusive).
Also `startTimestamp` is now guaranteed to be included and `endTimestamp` excluded.

The main changes are in `events.ts` and its unit tests. The rest is adapting existing code to call the new method.

Part of ENG-365

### DIFF
--- a/scripts/calculateRevenue/protocols/utils/events.ts
+++ b/scripts/calculateRevenue/protocols/utils/events.ts
@@ -117,9 +117,9 @@ export async function getBlockRange({
   ])
 
   // Validate the calculated block range.
-  // The startBlock must be strictly less than the endBlock for a valid, non-empty range.
-  // If startBlock == endBlock, the range is empty (e.g., startTimestamp and endTimestamp map to the same block for their >= condition).
-  // If startBlock > endBlock, it implies an issue, possibly with startTimestamp mapping to a block after endTimestamp's mapped block,
+  // The startBlock must be strictly less than the endBlockExclusive for a valid, non-empty range.
+  // If startBlock == endBlockExclusive, the range is empty (e.g., startTimestamp and endTimestamp map to the same block for their >= condition).
+  // If startBlock > endBlockExclusive, it implies an issue, possibly with startTimestamp mapping to a block after endTimestamp's mapped block,
   // though the initial timestamp check should largely prevent this specific sequence.
   if (startBlock >= endBlockExclusive) {
     throw new Error(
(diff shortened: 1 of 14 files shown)


scripts/calculateRevenue/protocols/utils/events.ts
```diff
@@ -42,13 +42,94 @@ export async function _getNearestBlock(
...
+  networkId,
+  startTimestamp,
+  endTimestamp,
+}: {
+  networkId: NetworkId
+  startTimestamp: Date // inclusive
+  endTimestamp: Date // exclusive
+}): Promise<{
+  startBlock: number // inclusive
+  endBlockExclusive: number
```
jeanregisser, line 101: Landed on this name for maximum clarity given the different behaviors of the underlying APIs.

scripts/calculateRevenue/protocols/utils/events.ts
```diff
@@ -42,13 +42,94 @@ export async function _getNearestBlock(
...
+ *                is *exclusive* from the desired range. When used in loops like
+ *                `for (let i = startBlock; i < endBlock; i++)`, or as an exclusive
+ *                upper bound in queries, it correctly defines the desired time window.
+ * @throws Will throw an error if startTimestamp is not before endTimestamp, or if a valid
+ *         block range cannot be determined (e.g., startBlock ends up >= endBlock).
+ */
+export async function getBlockRange({
+  networkId,
+  startTimestamp,
+  endTimestamp,
```
jeanregisser, line 94: I'm considering renaming this to `endTimestampExclusive` for maximum clarity too. In a follow up PR.
What do you think?

kathaypacific: sounds consistent! 👍🏻

kathaypacific (approved):
🥳 yay great!!!

scripts/calculateRevenue/protocols/utils/events.ts
```diff
@@ -75,25 +156,34 @@ export async function _fetchEvents({
...
   const events = []
 
-  while (currentBlock <= endBlock) {
-    const toBlock = Math.min(currentBlock + blocksPer, endBlock)
+  // Loop from startBlock up to (but not including) endBlock
+  while (currentBlock < endBlockExclusive) {
+    // The toBlock for getContractEvents is inclusive.
+    // So, it should be currentBlock + blocksPerQuery - 1, but not exceeding endBlock - 1.
+    const toBlockForQuery = Math.min(
+      currentBlock + blocksPerQuery - 1,
```
bakoushin, line 173: do we actually need this `-1` here?

scripts/calculateRevenue/protocols/utils/events.ts
bakoushin: i really like the explicit naming!

scripts/calculateRevenue/protocols/utils/events.ts
jeanregisser: It's not strictly needed, but more correct so we actually query the 10000 blocks, previously  it's was doing 10001

PR #142, opened 2025-05-14
chore: another divvi integrator reward processing

silasbw:
```
START=1745107200000
END=1747243524718
yarn ts-node ./scripts/getDivviIntegrators.ts \
--start-timestamp=$START \
--end-timestamp=$END \
--safe-transactions-file=rewards/divvi-integrators/${START}-${END}/safe-transactions.json
```

### DIFF
--- a/rewards/divvi-integrators/1745107200000-1747247130000/safe-transactions.json
+++ b/rewards/divvi-integrators/1745107200000-1747247130000/safe-transactions.json
@@ -1,0 +1,36 @@
+{
+  "meta": {},
+  "transactions": [
+    {
+      "to": "0x326161d68c05bE55367a0041b9C8f68082C04863",
+      "value": "0",
+      "data": null,
+      "contractMethod": {
+        "inputs": [
+          {
+            "internalType": "address[]",
+            "name": "users",
+            "type": "address[]"
+          },
+          {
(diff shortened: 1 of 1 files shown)


finnian0826 (approved):

rewards/divvi-integrators/1745107200000-1747243524718/safe-transactions.json
```diff
@@ -0,0 +1,36 @@
...
+            "type": "uint256[]"
+          }
+        ],
+        "name": "addRewards",
+        "payable": false
+      },
+      "contractInputsValues": {
+        "users": "[0xb82896c4f251ed65186b416dbdb6f6192dfaf926]",
+        "amounts": "[200000000]",
+        "rewardFunctionArgs": "[1745107200000, 1747243524718]"
```
jeanregisser, line 32: Because the underlying block timestamp precision is up to a second, we should preferably call the script with 0 milliseconds.
```suggestion
        "rewardFunctionArgs": "[1745107200000, 1747243524718]"
```

silasbw: ah, k, thanks, i'll update the internal docs too.

PR #143, opened 2025-05-14
feat: Add rhino protocol

finnian0826:
Gets all bridges a user completed between start and end timestamps, looks at the "from" token and converts amount to USD and sums for total amount.

Currently only fetches prices for tokens supported in Valora address-metadata, will create a follow up to get pricing for all tokens (supported by coingecko).

Tested with the following params:
```
const address = '0xAe30f674684C78C3c33A0e3c07be5ebfce25aD99'
const startTimestamp = new Date('2025-05-14T00:00:00.000Z')
const endTimestamp = new Date('2025-05-19T21:07:00.000Z')
```
and correctly found the 1 bridge that I completed on arbitrum to base for 1.94902347 USDC, correctly returns 1.94902347 as return value.


scripts/calculateRevenue/protocols/rhino/constants.ts
```diff
@@ -0,0 +1,25 @@
+import { Address, encodeEventTopics } from 'viem'
+import { NetworkId } from '../../../types'
+import { rhinoFiBridgeAbi } from '../../../abis/RhinoFiBridge'
+
+export const BRIDGE_VOLUME_USD_PRECISION = 8
+
+// This is what Rhino.fi uses in their event data to indicate the native token
+export const ALL_ZEROES_ADDRESS = '0x0000000000000000000000000000000000000000'
```
kathaypacific, line 8: `viem` has an export called `zeroAddress`, can we use that instead?

scripts/calculateRevenue/protocols/rhino/constants.ts
```diff
@@ -0,0 +1,25 @@
...
+export const NATIVE_TOKEN_DECIMALS = 18n
+
+// Contract addresses on different chains
+export const NETWORK_ID_TO_BRIDGE_CONTRACT_ADDRESS: Partial<
+  Record<NetworkId, Address>
+> = {
+  [NetworkId['base-mainnet']]: '0x2f59E9086ec8130E21BD052065a9E6B2497bb102',
+  [NetworkId['polygon-pos-mainnet']]:
+    '0xBA4EEE20F434bC3908A0B18DA496348657133A7E',
+  [NetworkId['celo-mainnet']]: '0x5e023c31E1d3dCd08a1B3e8c96f6EF8Aa8FcaCd1',
```
kathaypacific, line 18: should we use lower case for these addresses?

scripts/utils/index.ts
```diff
@@ -148,6 +148,35 @@ export const getBlock = memoize(_getBlock, {
   hash: (...params: Parameters<typeof _getBlock>) => params.join(','),
 })
 
+async function _getBlockNumber(
```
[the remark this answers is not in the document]
kathaypacific: we have `getFirstBlockAtOrAfterTimestamp` which isn't exported yet but i think it'd be preferrable to use that

scripts/calculateRevenue/protocols/rhino/index.ts
```diff
@@ -0,0 +1,177 @@
...
+}: {
+  address: string
+  startTimestamp: Date
+  endTimestamp: Date
+}): Promise<number> {
+  if (!isAddress(address)) {
+    throw new Error('Invalid address')
+  }
+  let totalRevenue = 0
+  for (const [networkId, contractAddress] of Object.entries(
```
kathaypacific, line 156: i wondered if we should parallelise the calculations across the networks, did you intend do to them in series to make it easier to diagnose errors?

scripts/calculateRevenue/protocols/rhino/index.ts
```diff
@@ -0,0 +1,177 @@
...
+  address: Address
+  contractAddress: Address
+  startTimestamp: Date
+  endTimestamp: Date
+  client: HypersyncClient
+  networkId: NetworkId
+}): Promise<BridgeTransaction[]> {
+  const fromBlock = await getBlockNumber(
+    networkId,
+    startTimestamp.getTime() / 1000, // Unix timestamp in seconds
```
kathaypacific, line 39: i think we might need to round this to the nearest integer if using `getFirstBlockAtOrAfterTimestamp`

scripts/calculateRevenue/protocols/rhino/index.ts
```diff
@@ -0,0 +1,177 @@
...
+        const hexData = bridge.data.startsWith('0x')
+          ? bridge.data.slice(2)
+          : bridge.data
+        // Check that the bridge is from the provided address (first block of data is sender)
+        if (
+          `0x${hexData.slice(24, 64)}`.toLowerCase() === address.toLowerCase()
+        ) {
+          const block = await getBlock(networkId, BigInt(bridge.blockNumber))
+          const blockTimestampDate = new Date(Number(block.timestamp) * 1000)
+          // And that the bridge happened within the time window
```
kathaypacific, line 65: if we already specify the from and end block in the hypersync query, does it not guarantee that the bridge happened in the time window?

scripts/calculateRevenue/protocols/rhino/index.ts
```diff
@@ -0,0 +1,177 @@
...
+    startTimestamp.getTime() / 1000, // Unix timestamp in seconds
+  )
+  const query = {
+    logs: [
+      { address: [contractAddress], topics: [[BRIDGED_DEPOSIT_WITH_ID_TOPIC]] },
+    ],
+    fieldSelection: {
+      log: [LogField.BlockNumber, LogField.Data],
+    },
+    fromBlock,
```
kathaypacific, line 48: should we also add the endBlock that matches the endTimestamp of the reward period?

scripts/calculateRevenue/protocols/rhino/index.ts
```diff
@@ -0,0 +1,177 @@
...
+      const tokenPrices = await fetchTokenPrices({
+        tokenId,
+        startTimestamp,
+        endTimestamp,
+      })
+      const tokenPriceUsd = getTokenPrice(tokenPrices, bridge.timestamp)
+      const partialUsdContribution =
+        Number(
+          (bridge.amount *
+            BigInt(tokenPriceUsd * 10 ** BRIDGE_VOLUME_USD_PRECISION)) /
```
kathaypacific, line 127: could we comment what `BRIDGE_VOLUME_USD_PRECISION` is? naively i would have expected the usd contribution to be `bridge.amount / 10^decimals * tokenPriceUsd`...

kathaypacific (commented):
looks great!

scripts/calculateRevenue/protocols/rhino/index.ts
finnian0826: The value is both multiplied and divided by `10 ** BRIDGE_VOLUME_USD_PRECISION`, I think this was more relevant back when we were doing revenue calculation like "1% of all swap fees" where the value would be small, this would be to preserve the decimals (8 in this case). Probably not necessary anymore unless the users are bridging very small amounts.

scripts/calculateRevenue/protocols/rhino/index.ts
finnian0826: looks like it takes a `Date` so was able to just pass in `startTimestamp`

scripts/calculateRevenue/protocols/rhino/index.ts
finnian0826: yes that should filter so I removed it. With past ones we didn't pass start/end so we needed that check but I think makes more sense and is more efficient to pass in the query.

scripts/calculateRevenue/protocols/rhino/index.ts
finnian0826: added! `toBlock` in hypersync is exclusive, so I could just re-use `getFirstBlockAtOrAfterTimestamp`

scripts/calculateRevenue/protocols/rhino/index.ts
finnian0826: I think makes sense to parallelize, added in [61cc9d7](https://github.com/divvi-xyz/divvi-protocol-v0/pull/143/commits/61cc9d7baa338f04498b8474367f03088250a7c4)

scripts/calculateRevenue/protocols/rhino/index.ts
finnian0826: Is there a precision we want in the revenue calculation? If so I can adjust the `BRIDGE_VOLUME_USD_PRECISION` value, otherwise can remove it if the math is okay without it.

finnian0826:
Looking more into it I think a person could call the deposit function but not have it actually complete a bridge. So changed it to looking at the withdrawal function which should better show when a bridge is complete. Manually tested again and worked as expected.

scripts/calculateRevenue/protocols/rhino/index.ts
```diff
@@ -0,0 +1,179 @@
...
+  if (userBridges.length === 0) {
+    return 0
+  }
+
+  let totalUsdContribution = 0
+
+  // For each bridge compute the USD contribution and add to the total
+  for (const bridge of userBridges) {
+    // Get the token decimals
+    const isNative = bridge.tokenAddress === zeroAddress
```
kathaypacific, line 101: if you have a link to some code in rhino that enforces this, it'd be helpful to add that as a comment :) i've usually see dapps use the 0xeeeeeee... address for native tokens, and use the zero address for minting and burning so i just found this convention to be a bit odd

scripts/calculateRevenue/protocols/rhino/index.ts
kathaypacific: okay, i see. we have the bignumber library in the project dependencies, if we convert `bridge.amount` and `tokenPriceUsd` to BigNumber instances, then i think we can avoid playing tricks to preserve precision and make this easier to read?

my brain is bad with simple math but naively it also seems that if we need a precision scaling factor, that we should apply it to `totalUsdContribution` after the summing, rather than to each `partialUsdContribution` which could be rounded to 0?

scripts/calculateRevenue/protocols/rhino/index.ts
```diff
@@ -0,0 +1,179 @@
...
+}: {
+  address: string
+  startTimestamp: Date
+  endTimestampExclusive: Date
+}): Promise<number> {
+  if (!isAddress(address)) {
+    throw new Error('Invalid address')
+  }
+
+  const totalRevenue = (
```
kathaypacific, line 150:
```suggestion
  const totalRevenueUsd = (
```

scripts/calculateRevenue/protocols/rhino/index.test.ts
```diff
@@ -0,0 +1,216 @@
...
+    expect(result[0].timestamp).toEqual(new Date('2025-01-01T23:30:00.000Z'))
+    expect(result[1].timestamp).toEqual(new Date('2025-01-02T23:30:00.000Z'))
+  })
+})
+
+describe('getTotalRevenueUsdFromBridges', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+  it('should return the correct total revenue in USD', async () => {
```
kathaypacific, line 149: if we are concerned about precision loss, should we update this test to make the token price and bridge amount really small so that we also cover that scenario?

kathaypacific (approved):
🚀

jeanregisser (commented):
Great! 👍

scripts/calculateRevenue/protocols/rhino/index.test.ts
finnian0826: I think that was more of a concern for protocols with small values, like gas fees or some small percent of transaction value. So I think we are okay without it for here.

PR #144, opened 2025-05-14
chore: If no allowlist passed in to celoTransactions, accept all referrals

finnian0826:


scripts/protocolFilters/celoTransactions.ts
```diff
@@ -7,8 +7,9 @@ export async function filter(
   event: ReferralEvent,
   builderAllowList?: Address[],
 ): Promise<boolean> {
-  const allAllowlistedBuilders = KNOWN_BUILDERS.concat(builderAllowList ?? [])
-  return allAllowlistedBuilders.some(
-    (address) => address.toLowerCase() === event.referrerId.toLowerCase(),
-  )
+  return builderAllowList
```
silasbw, line 10: I think this is easier to reason/read, but not a huge deal:

```ts
if (!builderAllowList) {
  return true
}
```

silasbw (approved):

PR #145, opened 2025-05-15
refactor: remove input and output file name args for scripts

kathaypacific:
We need to run 3 scripts to do the reward distributions, and each script requires us to declare the input and output file paths. Each file path depends on the reward period start and end times. This is cumbersome and there is room for copy and paste errors when creating the file paths.

This PR proposes the following changes:
- we pass the start and end timestamp to all scripts (currently not needed for `fetchReferrals`) - this might make it easy to eventually automate and declare these variables one time?
- the input and output file paths are determined in code and use the start and end timestamps + protocol
- the folder names will look like `rewards/{{protocol}}/2025-05-08T00-00-00Z_2025-05-15T00-00-00Z` which i personally find more readable than 2 unix timestamps, and still allows sorting alphabetically + chronologically
- the file names will continue to be `referrals.csv`, `revenue.csv`, `safe-transactions.json`
- a sequence of reward distribution commands will look like (for celo PoI):
```
yarn ts-node scripts/fetchReferrals.ts --protocol=celo-transactions --start-timestamp=2025-05-08T00:00:00Z --end-timestamp=2025-05-15T00:00:00Z --builder-allowlist-file=builder-allowlist.csv
yarn ts-node scripts/calculateRevenue.ts --protocol=celo-transactions --start-timestamp=2025-05-08T00:00:00Z --end-timestamp=2025-05-15T00:00:00Z
yarn ts-node scripts/calculateRewards/proofOfImpact.ts --start-timestamp=2025-05-08T00:00:00Z --end-timestamp=2025-05-15T00:00:00Z
```

The downside is that the scripts expect certain input files to exist and we are less free to run the scripts in isolation, but I think this is not a real scenario that is worth optimising for.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,6 +120,3 @@ gha-creds-*.json
 # solidity-coverage files
 /coverage
 /coverage.json
-
-# CSV Output files at the root
-*.csv
--- a/scripts/calculateRevenue.ts
+++ b/scripts/calculateRevenue.ts
@@ -1,23 +1,33 @@
 import calculateRevenueHandlers from './calculateRevenue/protocols'
 import { parse } from 'csv-parse/sync'
 import { stringify } from 'csv-stringify/sync'
-import { readFileSync, writeFileSync } from 'fs'
+import { mkdirSync, readFileSync, writeFileSync } from 'fs'
 import yargs from 'yargs'
-import { protocols, Protocol } from './types'
+import { protocols } from './types'
+import { toPeriodFolderName } from './utils/dateFormatting'
+import { dirname } from 'path'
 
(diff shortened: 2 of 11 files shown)


.gitignore
```diff
@@ -120,6 +120,3 @@ gha-creds-*.json
 # solidity-coverage files
 /coverage
 /coverage.json
-
-# CSV Output files at the root
-*.csv
```
kathaypacific, line 125: we want to check in the reward metadata i think

jeanregisser (approved):
Love it! 🚀

silasbw (approved):

PR #149, opened 2025-05-16
chore: use ISO 8601 timestamps for directories

silasbw:
A well-defined format is useful because it's parseable by js Date or GNU date. We also need more precision for some programs we're running, like Divvi integration rewards.


jeanregisser (approved):
LGTM 🚀

kathaypacific:
i wasn't sure if colons and periods were going to be an issue in filenames, but if not then this is ideal :D

PR #155, opened 2025-05-19
fix: use end timestamp for fetching referrals

kathaypacific:
As the title - fetch referrals currently returns all referrals until the time of script execution. This is unnecessary, because we only need to calculate rewards for referrals up to the reward end period. It also causes extra confusion when peer reviewing the reward distribution, since the output of the script is time dependent.

Also fixes referral buffer time incorrect conversion.

### DIFF
--- a/scripts/calculateRewards/proofOfImpact.ts
+++ b/scripts/calculateRewards/proofOfImpact.ts
@@ -36,20 +36,17 @@ export function calculateRewardsProofOfImpact({
   )
   const totalRewardsForPeriod = timeDiff.times(rewardsPerMillisecond)
 
-  const referrerKpis = kpiData
-    // filter out rows with no revenue, which is possible for users who were referred between the reward period end date and the time of the reward distribution
-    .filter((row) => BigInt(row.revenue) > 0)
-    .reduce(
-      (acc, row) => {
-        if (!(row.referrerId in acc)) {
-          acc[row.referrerId] = BigInt(row.revenue)
-        } else {
-          acc[row.referrerId] += BigInt(row.revenue)
-        }
-        return acc
-      },
-      {} as Record<string, bigint>,
-    )
+  const referrerKpis = kpiData.reduce(
+    (acc, row) => {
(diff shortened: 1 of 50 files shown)


scripts/calculateRewards/proofOfImpact.ts
```diff
@@ -36,20 +36,17 @@ export function calculateRewardsProofOfImpact({
   )
   const totalRewardsForPeriod = timeDiff.times(rewardsPerMillisecond)
 
-  const referrerKpis = kpiData
-    // filter out rows with no revenue, which is possible for users who were referred between the reward period end date and the time of the reward distribution
-    .filter((row) => BigInt(row.revenue) > 0)
```
kathaypacific, line 41: i realised this filter needs to go on all rewards, because it is always possible that the revenue calculation for a referred user is 0 (if a builder doesn't not attach the referral to a KPI transaction). this is only needed so that we don't post 0 rewards, so i moved it to the createSafeTransactionBatch function instead

scripts/fetchReferrals.ts
```diff
@@ -76,7 +78,7 @@ async function main() {
   const outputEvents = filteredEvents.map((event) => ({
     referrerId: event.referrerId,
     userAddress: event.userAddress,
-    timestamp: event.timestamp,
+    timestamp: new Date(event.timestamp * 1000).toISOString(),
```
kathaypacific, line 81: to make it easier to reason about the output file

scripts/utils/referrals.ts
```diff
@@ -37,6 +38,7 @@ export async function fetchReferralEvents(
   protocol: Protocol,
   referrerIds?: Address[],
   useStaging = false,
+  endTimestamp?: Date,
```
jeanregisser, line 41: nit: since I've just merged the rename [PR](https://github.com/divvi-xyz/divvi-protocol-v0/pull/154).
```suggestion
  endTimestampExclusive?: Date,
```

jeanregisser (approved):
🚀

PR #156, opened 2025-05-19
fix: referral buffer time incorrect conversion

kathaypacific:
A bug snuck through where we were creating a date with a timestamp in seconds, when it expects milliseconds. This caused the calculateRevenue script to calculate starting from the unix epoch instead of the reward period specified.

### DIFF
--- a/scripts/calculateRevenue.ts
+++ b/scripts/calculateRevenue.ts
@@ -8,7 +8,7 @@ import { toPeriodFolderName } from './utils/dateFormatting'
 import { dirname } from 'path'
 
 // Buffer to account for time it takes for a referral to be registered, since the referral transaction is made first and the referral registration happens on a schedule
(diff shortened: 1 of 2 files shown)


scripts/calculateRevenue.ts
```diff
@@ -41,12 +38,10 @@ async function main(args: ReturnType<typeof parseArgs>) {
       `Calculating revenue for ${userAddress} (${i + 1}/${eligibleUsers.length})`,
     )
 
-    const referralTimestamp = new Date(
-      timestamp - REFERRAL_TIME_BUFFER_IN_SECONDS,
-    )
+    const referralTimestamp = new Date(timestamp)
```
kathaypacific, line 41: this doesn't look like i've fixed anything! but in the base branch i proposed that `timestamp` is now a ISO date string, so that it is easier to reason about when reviewing the data

jophish (approved):

PR #161, opened 2025-05-19
chore: add a proportionalPrizeContest reward

silasbw:
This is the calculation we use for proof-of-impact, PG, and scout game. I'll update those to use this code in a follow-on.

### DIFF
--- a/scripts/calculateRewards/proportionalPrizeContest.ts
+++ b/scripts/calculateRewards/proportionalPrizeContest.ts
@@ -1,0 +1,50 @@
+import BigNumber from 'bignumber.js'
+
+interface KpiRow {
+  referrerId: string
+  userAddress: string
+  revenue: string
+}
+
+export function calculateProportionalPrizeContest({
+  kpiData,
+  rewards,
+}: {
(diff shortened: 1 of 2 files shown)


scripts/calculateRewards/proportionalPrizeContest.ts
```diff
@@ -0,0 +1,50 @@
...
+  const total = Object.values(referrerKpis).reduce(
+    (sum, value) => sum + value,
+    BigInt(0),
+  )
+
+  const rewardsPerReferrer = Object.entries(referrerKpis).map(
+    ([referrerId, kpi]) => {
+      return {
+        referrerId,
+        kpi,
```
finnian0826, line 40: Is the `kpi` field needed? I assume the output of this will be passed to `createAddRewardSafeTransactionJSON`, and the reward object in that function does not expect/use `kpi`.

(I ran into issues with this in testing for celo-pg because it was trying to serialize bigint, doesn't seem like issue here but wondering if it even needs to be here).

finnian0826 (approved):
looks good, just small question about potentially unused field

PR #162, opened 2025-05-19
chore: update proofOfImpact to call proportionalPrizeContest

silasbw:


finnian0826 (approved):

scripts/calculateRewards/proofOfImpact.ts
```diff
@@ -133,7 +106,6 @@ async function main(args: ReturnType<typeof parseArgs>) {
     'rewards:',
     rewards.map((r) => ({
       referrerId: r.referrerId,
-      kpi: r.kpi,
```
jeanregisser, line 136: nit: I liked having this info to as a quick sanity check to see if the reward looks plausible compared to the KPI.

PR #172, opened 2025-05-21
fix(scripts): use correct block ranges per chain in scoutGameV0 KPI calculation

jeanregisser:
Corrects an issue in scout game KPI calculation (spotted in https://github.com/divvi-xyz/divvi-protocol-v0/pull/159#discussion_r2099677928) where `getBlockRange` was only called for Celo. This fix ensures `getBlockRange` is called for each configured network.


bakoushin (approved):
👨🏻‍🔧

silasbw (approved):
oh man, :scream: !

PR #174, opened 2025-05-21
feat: calculate and upload kpis periodically

kathaypacific:
This PR sets up a CI that:
- runs on the hour every hour
- `fetchReferrals` and `calculateKpi` per campaign will be executed
- the resulting files are uploaded to google cloud storage (each upload replaces the previous upload, for the same reward period)

Some details and thoughts:
- it would be really nice to have the uploaded data to be the source of truth for the reward calculations so that the engineer doing the reward distribution does not need to run the same scripts locally (especially because they are getting slow). there is an edge case where the period data is incomplete if the last workflow of the period fails, we need to think about how to handle this.
- i intentionally kept the agreed upon file and folder structure, namely one kpi file per campaign per reward period. this means that the files for completed reward periods are "complete" and no longer updated, but the current reward period file is updated periodically (can think of it like, the filename mentions a date in the future, so the file is not yet "complete")

When adding a new campaign, we need to add the campaign and reward periods to the new campaign data object in the `uploadCurrentPeriodKpis` file. The campaign will then be included in subsequent executions.

An example of a [successful execution](https://github.com/divvi-xyz/divvi-protocol-v0/actions/runs/15184028178/job/42700141737), where the files were written to the divvi-staging [bucket](https://console.cloud.google.com/storage/browser/divvi-campaign-data?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&inv=1&invt=AbyDdA&project=divvi-staging). An example of a [successful execution](https://github.com/divvi-xyz/divvi-protocol-v0/actions/runs/15191866810/job/42726471647?pr=174) on production where the files were written to divvi-production [bucket](https://console.cloud.google.com/storage/browser/divvi-campaign-data-production).

Waiting for #176, #178, #179 to be merged first.

Related to ENG-392


.github/workflows/calculate-and-upload-KPIs.yaml
```diff
@@ -0,0 +1,36 @@
+name: Calculate and Upload KPIs
+
+on:
+  schedule:
+    - cron: '0 * * * *' # Runs at minute 0 of every hour
+  pull_request:
```
kathaypacific, line 6: i'll remove this, i wanted to test the workflow on this branch once

silasbw (approved):

PR #175, opened 2025-05-21
refactor: calculate kpi in batches

kathaypacific:
We have ~1.3k referred end users so far for a campaign, and it is taking [~1 second to calculate the kpi per user](https://github.com/divvi-xyz/divvi-protocol-v0/actions/runs/15163741596/job/42635973237?pr=174). This means ~20 minutes of processing, which is annoying for the reward distribution person but also limits how frequently we can display current kpi data. This PR parallelises the calculations per user.

I tested this with 100 users, it took ~6 seconds to do it batched and 28 seconds to do it in series. I was careful to use new users for testing, to minimise the chances of caching affecting the results.

Related to ENG-392


silasbw (approved):

silasbw:
glad you did this :sweat_smile: !

PR #178, opened 2025-05-22
chore: ensure delays to avoid rate limits in calculateKpi

kathaypacific:
When running the calculateKpi scripts on CI, it was the _getNearestBlock function that started failing. On closer inspection, it seems envio has no rate limits but DefiLlama has a rate limit of 500 requests/min. In the worst case, we call _getNearestBlock for every user in the reward period if they were referred within the period (i.e. we need to get the block corresponding to their referral)

This PR adds:
- better error messaging on _getNearestBlock (the error response was previously logged as `[object Object]`
- delay after every kpi batch transaction to keep requests under the rate limit 

Related to ENG-392

### DIFF
--- a/scripts/calculateKpi.ts
+++ b/scripts/calculateKpi.ts
@@ -3,7 +3,7 @@ import { parse } from 'csv-parse/sync'
 import { stringify } from 'csv-stringify/sync'
 import { mkdirSync, readFileSync, writeFileSync } from 'fs'
 import yargs from 'yargs'
-import { protocols } from './types'
+import { Protocol, protocols } from './types'
 import { toPeriodFolderName } from './utils/dateFormatting'
 import { dirname, join } from 'path'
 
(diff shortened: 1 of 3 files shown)


scripts/calculateKpi.ts
```diff
@@ -89,6 +93,11 @@ async function calculateKpiBatch({
         (result): result is NonNullable<typeof result> => result !== null,
       ),
     )
+
+    // Add a 1 minute delay to avoid rate limits from DefiLlama
+    if (i > 0) {
```
bakoushin, line 98: this condition will not delay after the first batch, right? so the second batch will not be delayed. 
why do we need that?

bakoushin (approved):

scripts/calculateKpi.ts
bakoushin: and `delayPerBatchInMs` seems not 1 minute. maybe the comment is stale

PR #179, opened 2025-05-22
refactor: return outputFile from getArgs

kathaypacific:
This change does not affect how we are calling the scripts from the CLI but enables us to specify the full output path when calling the function from another script (I want to be able to pass a custom filepath here when running the calculations from CI)

Related to ENG-392

### DIFF
--- a/scripts/fetchReferrals.ts
+++ b/scripts/fetchReferrals.ts
@@ -46,11 +46,19 @@ async function getArgs() {
       type: 'string',
     }).argv
 
+  const outputDir = join(
+    argv['datadir'],
+    argv['protocol'],
+    toPeriodFolderName({
+      startTimestamp: new Date(argv['start-timestamp']),
+      endTimestampExclusive: new Date(argv['end-timestamp']),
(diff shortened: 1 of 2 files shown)


scripts/fetchReferrals.ts
```diff
@@ -86,15 +94,7 @@ export async function fetchReferrals(
...
-  const outputDir = join(
-    args.datadir,
-    args.protocol,
-    toPeriodFolderName({
-      startTimestamp: new Date(args.startTimestamp),
-      endTimestampExclusive: new Date(args.endTimestampExclusive),
-    }),
-  )
-  const outputFile = `${outputDir}/referrals.csv`
+  const outputFile = `${args.outputDir}/referrals.csv`
```
bakoushin, line 97: I guess it's safer to always use `join` instead of `${dir}/${path}` to avoid any surprises with trailing slashes/different filesystems

kathaypacific: oh yes, updated! thanks :D

PR #182, opened 2025-05-22
feat: allow manual dispatch of data upload workflow

kathaypacific:
This PR adds the possibility to run the data upload CI manually against a specific timestamp. I think this would be useful if the last periodic calculation for a reward period fails, and we want to "complete" the data for that period. Or, before a distribution, we want to ensure that the data is complete by running the CI manually and using the uploaded data to calculate rewards. The calculations are taking like ~30 minutes now so it is nice not to have to do them on your local machine.

Related to ENG-392


jeanregisser (approved):
🚀

PR #192, opened 2025-05-28
fix: calculate gas used for celo-pg

silasbw:


satish-ravi (approved):

scripts/calculateKpi/protocols/utils/networks.ts
```diff
@@ -14,7 +14,7 @@ export async function fetchTotalTransactionFees({
   startBlock?: number // inclusive
   endBlockExclusive?: number
 }): Promise<number> {
-  let totalTransactionFees = 0
+  let totalGasUsed = 0
```
silasbw, line 17: not sure if this should be a bigint or not. seems better safe than sorry and I can look into using bigint in a separate PR.

PR #195, opened 2025-05-29
chore: update calculateKpi to use ResultDirectory

silasbw:


scripts/calculateKpi.ts
```diff
@@ -106,17 +102,14 @@ export async function calculateKpi(args: Awaited<ReturnType<typeof getArgs>>) {
   const endTimestampExclusive = new Date(args.endTimestampExclusive)
   const protocol = args.protocol
 
-  const inputFile = join(args.outputDir, 'referrals.csv')
-  const outputFile = join(args.outputDir, 'kpi.csv')
+  const resultDirectory = new ResultDirectory({
+    datadir: args.datadir,
+    name: protocol,
+    startTimestamp,
+    endTimestampExclusive,
```
finnian0826, line 109: This is the timestamp to calculate up to (based on when the workflow runs), but it should be the end timestamp of the campaign. Fixed in https://github.com/divvi-xyz/divvi-protocol-v0/pull/197.

PR #204, opened 2025-05-31
chore(celo-pg): option to split reward amount

silasbw:


jophish (approved):
math lgtm, A’d assuming checks pass. what’s the context for this change however? to slightly reduce the effect of outliers?

PR #209, opened 2025-06-02
chore: document calculateKpi functions

jeanregisser:
### Description

Adds comprehensive TypeDoc comments to all `calculateKpi` functions (11 protocols) with standardized documentation including:
- **KPI Unit & Business Purpose**: What each metric measures
- **Data Sources**: Actual technologies used (HyperSync, RPC, The Graph, APIs)  
- **Calculation Method**: Step-by-step methodology
- **Business Assumptions**: Revenue models and attribution logic

The goal is to make KPI calculation logic more transparent and maintainable.

### Feedback Requested 🙏

For folks who implemented these KPI calculation functions, please review to ensure:
- Business assumptions are accurate
- Calculation methods match implementation
- No missing protocol-specific details

### Disclaimer

*AI assistance was used to generate this documentation. Human review is essential for accuracy.*

Part of ENG-394


scripts/calculateKpi/protocols/aave/index.ts
```diff
@@ -163,6 +297,18 @@ function splitUserEarningHistoryIntoSegments(
   })
 }
 
+/**
+ * Converts reserve factor history into time-segmented data.
+ *
+ * **Business Purpose**: Creates discrete time segments showing which reserve factor
+ * was active during each period, enabling precise protocol revenue calculations.
+ *
+ * **Reserve Factor**: Percentage of borrower interest that goes to protocol (typically 10-25%)
```
bakoushin, line 306: Maybe we can swap "typically" for "for example 20%" because it may change in the future.

scripts/calculateKpi/protocols/aave/index.ts
```diff
@@ -17,17 +23,83 @@ interface RevenueCalculationContext {
...
+ * and borrowing activity on Aave protocol. This metric quantifies the economic value generated by user
+ * interactions with Aave's money markets and supports protocol revenue analysis.
+ *
+ * **Protocol Context**: Aave is a decentralized lending protocol that allows users to lend and borrow
+ * cryptocurrencies. It generates revenue through interest rate spreads and various fees, with revenue
+ * allocation determined by reserve factors set per asset market.
+ *
+ * **Supported Networks**: Multiple EVM-compatible networks where Aave has deployed lending pools
+ *
+ * **Reserve Factor Model**: Each asset market has a specific reserve factor (0-30%) that determines
```
bakoushin, line 65: it can be 50% (currently for WBTC) or even 80%.
probably it is safer to say "for example, 20%".

scripts/calculateKpi/protocols/aave/index.ts
```diff
@@ -17,17 +23,83 @@ interface RevenueCalculationContext {
...
+ *
+ * **Data Sources**:
+ * - **The Graph**: Aave subgraphs for lending/borrowing events, reserve data, and interest calculations
+ * - **RPC Queries**: Reserve factors and pool configurations via Viem public client calls
+ * - **Token Price API**: Historical token prices via `fetchTokenPrices` utility for USD conversion
+ * - **Block Data**: Timestamps via `getBlockRange` utility for temporal filtering
+ *
+ * **Business Assumptions**:
+ * - Revenue is derived from interest rate spreads and reserve factors applied to borrowing activity
+ * - User's revenue contribution is proportional to their lending/borrowing positions within time window
```
bakoushin, line 76: "lending/borrowing positions" -> "deposits"

scripts/calculateKpi/protocols/aave/index.ts
```diff
@@ -17,17 +23,83 @@ interface RevenueCalculationContext {
...
+ * - USD conversion uses token prices at time of each transaction for accuracy
```
bakoushin, line 77: we don't convert each transaction, only the total amounts when we have to sum up the revenue denominated in different tokens. probably better "at the time of the revenue calculation"

scripts/calculateKpi/protocols/aave/index.ts
```diff
@@ -17,17 +23,83 @@ interface RevenueCalculationContext {
...
+ * - Reserve factors are applied consistently across all supported markets
+ * - Interest calculations account for compound interest and variable rate changes
+ *
+ * **Revenue Sources**:
+ * - Interest rate spreads between lending and borrowing rates
```
bakoushin, line 82: again, borrowing seems redundant here

scripts/calculateKpi/protocols/aave/index.ts
```diff
@@ -17,17 +23,83 @@ interface RevenueCalculationContext {
...
+ * - Reserve factors applied to total protocol interest
+ * - Liquidation fees and flash loan fees
+ * - Protocol-specific fees (varies by network and governance decisions)
+ *
+ * **Calculation Method**:
+ * 1. Queries user's lending and borrowing positions across all supported networks
```
bakoushin, line 88: leading and borrowing positions -> deposits

scripts/calculateKpi/protocols/aave/index.ts
```diff
@@ -17,17 +23,83 @@ interface RevenueCalculationContext {
...
+ * 2. Retrieves interest accrual events and reserve factor data for each position
```
bakoushin, line 89: "Retrieves interest accrual events and reserve factor data" -> "Calculates user earnings and retrieves reserve factor data"

scripts/calculateKpi/protocols/aerodrome/index.ts
```diff
@@ -4,6 +4,56 @@ import {
 } from './constants'
 import { calculateRevenueDrome } from '../utils/drome/calculateRevenueDrome'
 
+/**
+ * Calculates trading fee revenue generated for Aerodrome DEX on Base Network.
```
finnian0826, line 8: Note: For aerodrome and velodrome, it only gets the trading fee for the final transfer in the swap route. I.e., if it is user -> pool 1 -> pool 2 -> user, only the pool 2 -> user fee is included.

scripts/calculateKpi/protocols/fonbnk/index.ts
```diff
@@ -13,6 +13,30 @@ import { getTokenPrice } from '../beefy'
 import { FonbnkTransaction, SUPPORTED_FONBNK_NETWORKS } from './types'
 import { Address, fromHex, isAddress, pad } from 'viem'
 
+/**
+ * Retrieves all token transfer transactions from Fonbnk payout wallets to a specific user.
```
finnian0826, line 17: Note: only for cash-in with Fonbnk, I think that is captured by saying "payout wallets to user..."

scripts/calculateKpi/protocols/velodrome/index.ts
```diff
@@ -4,6 +4,53 @@ import {
...
+ * - Trading fees are proportional to trade volume and pool-specific fee rates
+ * - Revenue attribution is based on user's direct trading activity within specified time window
+ * - USD conversion uses token prices at time of each transaction for accuracy
+ * - Only whitelisted liquidity pools are included in revenue calculations
+ * - Fee rates are dynamically set per pool based on volatility and governance decisions
+ *
+ * **Shared Implementation**: Uses the common `calculateRevenueDrome` utility which handles the core
+ * trading fee calculation logic shared between Velodrome and Aerodrome (fork relationship).
+ *
+ * **Calculation Method**:
```
finnian0826, line 41: This seems to differ a bit from the aerodrome one where it doesn't mention `calculateRevenueDrome` and just gives the steps done instead, maybe should make them consistent since they are essentially the same just with different input params?

finnian0826 (approved):

jh2oman (approved):
I updated the somm docs directly and it looks good now!

kathaypacific (approved):
LGTM!

scripts/calculateKpi/protocols/beefy/index.ts
```diff
@@ -126,6 +164,53 @@ export async function calculateVaultRevenue(
   return totalUsdContribution
 }
 
+/**
+ * Calculates vault management fee revenue generated for Beefy Finance protocol.
+ *
+ * **KPI Unit**: USD (United States Dollars)
+ *
+ * **Business Purpose**: Measures the vault management fee revenue attributable to a specific user's
+ * deposit activity in Beefy vaults. This metric quantifies the economic value generated by user
```
jophish, line 173:
```suggestion
 * deposit activity in Beefy vaults over a given time span. This metric quantifies the economic value generated by user
```

scripts/calculateKpi/protocols/beefy/index.ts
```diff
@@ -126,6 +164,53 @@ export async function calculateVaultRevenue(
...
+ *
+ * **Data Sources**:
+ * - **Beefy API**: Vault configurations, TVL data, and strategy metadata from `api.beefy.finance`
+ * - **RPC Queries**: User deposit/withdrawal events via Viem public client calls to vault contracts
+ * - **Token Price API**: Historical token prices via `fetchTokenPrices` utility for USD conversion
+ * - **Block Data**: Timestamps via `getBlockRange` utility for temporal filtering
+ * - **Network RPCs**: Multi-network vault contract interactions across Arbitrum, Polygon, Base, Optimism
+ *
+ * **Business Assumptions**:
+ * - Standard management fee rate is 4.5% of total vault yield
```
jophish, line 190: This is not an assumption that is baked into code anywhere, this could be a bit misleading.

scripts/calculateKpi/protocols/beefy/index.ts
```diff
@@ -126,6 +164,53 @@ export async function calculateVaultRevenue(
...
+ * - Revenue attribution is proportional to user's share of vault deposits within time window
+ * - USD conversion uses token prices at time of each transaction
+ * - Only active vault strategies with positive TVL are included
```
jophish, line 193: Is this true? We iterate over all vaults that a user has ever interacted with, so this is not necessarily true in some cases. E.g., if a user has funds in a vault, that vault is then shut down, and then we run the script over a time range where the user _did_ have funds in the vault, we'll include an "inactive" vault strategy.

scripts/calculateKpi/protocols/beefy/index.ts
```diff
@@ -126,6 +164,53 @@ export async function calculateVaultRevenue(
...
+ * - Fee distribution is immediate and based on real-time vault performance
+ *
+ * **Fee Structure**: 4.5% performance fee on all vault yields (industry standard for yield aggregators)
```
jophish, line 196: Again, this isn't baked into the code anywhere if I recall correctly; the performance fee is only ever _implied_ by the relationship between absolute fee amount and total vault TVL -- we never actuallly care about the performance fee percentage explicitly in the script.

scripts/calculateKpi/protocols/beefy/index.ts
```diff
@@ -126,6 +164,53 @@ export async function calculateVaultRevenue(
...
+ *
+ * **Calculation Method**:
+ * 1. Retrieves active vault configurations from Beefy API
```
jophish, line 199:
```suggestion
 * 1. Retrieves all configurations for vaults with which user has had transactions from Beefy API
```

scripts/calculateKpi/protocols/beefy/index.ts
```diff
@@ -126,6 +164,53 @@ export async function calculateVaultRevenue(
...
+ * 2. Filters vaults by supported networks and active status
+ * 3. For each vault, queries user's deposit/withdrawal events within time window
```
jophish, line 201:
```suggestion
 * 3. For each vault, queries user's deposit/withdrawal events within time window
 * 4. For each vault, queries historical Fee events within time window
 * 5. Calculates user's proportional share of vault deposits at the time of each Fee event
 * 6. Determines user contribution to management fee based off their proportional share of vault TVL
 * 7. Converts to USD using historical token prices at transaction timestamps
 * 8. Aggregates user fee contribution across all vaults and networks to get total user fee contribution
 ```

This is from memory, but I believe this is more correct.

PR #211, opened 2025-06-02
feat: add lisk campaign kpi calcalculation

kathaypacific:
I did a small manual test to check that the kpi calculation works e2e, I used a random address on the Lisk block explorer and verified that the gas calculated by hypersync per tx matches what is on the explorer.

Related to ENG-415


scripts/calculateKpi/protocols/liskV0/index.ts
```diff
@@ -0,0 +1,26 @@
+import { NetworkId } from '../../../types'
+import { getBlockRange } from '../utils/events'
+import { fetchTotalGasUsed } from '../utils/networks'
+
+export async function calculateKpi({
```
kathaypacific, line 5: this is the same content as the celo-pg calculation, with a different network

scripts/protocolFilters/lisk-v0.ts
```diff
@@ -0,0 +1,5 @@
+import { MatcherFn } from '../types'
+
+export const filter: MatcherFn = async (event) => {
+  return !!event
+}
```
kathaypacific, line 5: these filters are a bit boilerplate, would be nice to remove.....?

bakoushin (approved):
🚀

PR #214, opened 2025-06-03
chore: move scout game v0 to use ResultDirectory

silasbw:


finnian0826 (approved):

scripts/calculateRewards/scoutGameV0.ts
```diff
@@ -67,32 +64,18 @@ function parseArgs() {
...
-    args.datadir,
-    'scout-game-v0',
-    toPeriodFolderName({
-      startTimestamp,
-      endTimestampExclusive,
-    }),
-  )
-  const inputPath = join(folderPath, 'kpi.csv')
-  const outputPath = join(folderPath, 'safe-transactions.json')
+  const resultDirectory = new ResultDirectory({
```
finnian0826, line 71: Do we want this to be an arg instead so we can just pass it in to main?

PR #216, opened 2025-06-03
chore: calculateSqrtProportionalPrizeContest

silasbw:


scripts/calculateRewards/proportionalPrizeContest.ts
```diff
@@ -47,3 +47,50 @@ export function calculateProportionalPrizeContest({
 
   return rewardsPerReferrer
 }
+
+export function calculateSqrtProportionalPrizeContest({
```
silasbw, line 51: i spend too long futzing with different APIs to avoid code duplication with the function above, but it wasn't obvious what the right API/erogonmics was, so I ended up with this.  we can update it once we know (or if we add more  options for the proportional prize, etc).

scripts/calculateRewards/proportionalPrizeContest.ts
```diff
@@ -47,3 +47,50 @@ export function calculateProportionalPrizeContest({
...
+        acc[row.referrerId] += BigInt(row.kpi)
+      }
+      return acc
+    },
+    {} as Record<string, bigint>,
+  )
+
+  const referrerPowerKpis = Object.entries(referrerKpis).reduce(
+    (acc, [referrerId, kpi]) => {
+      acc[referrerId] = BigNumber(kpi).sqrt()
```
kathaypacific, line 72: random thought, is it likely we'd want to play with the power? i wonder if we should pass in a number (0.5 for sqrt) and use `.pow` since the variable name mentions power 🤷🏻‍♀️

kathaypacific (approved):
LGTM!

finnian0826 (approved):
lgtm! In a follow up this will be used in scout-game-v0 reward calculation right? And also KPI added to the return value, that way it can be used in the dashboard. Just checking to make sure I'm setting up dashboard correctly!

silasbw:
Roger @finnian0826 !

PR #220, opened 2025-06-04
feat: improve DefiLlama rate limit handling

kathaypacific:
This PR adds the `bottleneck` npm package to manage the DefiLlama rate limit at the api request level. This is much more efficient than managing at the kpi calc level, since some calculations don't require a request to DefiLlama.

Example of a [job](https://github.com/divvi-xyz/divvi-protocol-v0/actions/runs/15451499250/job/43494472674?pr=220) running with this change, which took ~15 minutes. It ran the same logic as the other jobs in the CI which take ~90 minutes.

With this change, I could also change the frequency of the scheduled calc and upload job back to 1 hour.

Related to ENG-414

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "@typescript-eslint/parser": "^8.33.0",
     "@valora/prettier-config": "^0.0.1",
     "bignumber.js": "^9.2.1",
+    "bottleneck": "^2.19.5",
     "chai": "^4.5.0",
     "csv-parse": "^5.6.0",
     "csv-stringify": "^6.5.2",
--- a/.github/workflows/calculate-and-upload-KPIs.yaml
+++ b/.github/workflows/calculate-and-upload-KPIs.yaml
@@ -28,7 +28,8 @@ jobs:
       - name: Calculate and upload KPIs for the current reward period
         run: |
           yarn ts-node scripts/uploadCurrentPeriodKpis.ts \
-            --calculation-timestamp ${{ github.event.inputs.timestamp }}
+            --calculation-timestamp ${{ github.event.inputs.timestamp }} \
+            --dry-run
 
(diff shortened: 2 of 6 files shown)


finnian0826 (approved):
LGTM! Is there a reason that we want this every hour vs. every 2 hours? I'm just thinking if we have a lot of campaigns going at the same time the time for the upload script might approach an hour again. And it seems like this data is just used in the Dune dashboard, not sure how fresh we want that data (or even how fresh it is since queries would need to be re-run to update every hour I think).

silasbw (approved):
nice

package.json
```diff
@@ -52,6 +52,7 @@
     "@typescript-eslint/parser": "^8.33.0",
     "@valora/prettier-config": "^0.0.1",
     "bignumber.js": "^9.2.1",
+    "bottleneck": "^2.19.5",
```
jeanregisser, line 55: nit: I see this lib hasn't been updated in 6 years, maybe we can switch with another one?
quick search yielded https://github.com/jhurliman/node-rate-limiter

jeanregisser (approved):
Awesome! 🚀

kathaypacific:
> LGTM! Is there a reason that we want this every hour vs. every 2 hours? I'm just thinking if we have a lot of campaigns going at the same time the time for the upload script might approach an hour again. And it seems like this data is just used in the Dune dashboard, not sure how fresh we want that data (or even how fresh it is since queries would need to be re-run to update every hour I think).

i was thinking that more up to date data is better, like in an ideal world you'd want to access live data. we're not there yet but i think reducing the interval between calculations is inherently good for whoever is viewing the data. i have some plans to add a redis cache to this setup so it should speed up even more and i hope that we don't have to worry about this job running for longer than a few minutes :)

PR #221, opened 2025-06-05
fix: lower defiLlama rate limit

kathaypacific:
Last time I was researching, I thought that I saw 500/min as the rate limit. But according to [this page](https://docs.llama.fi/pro-api#open-free), it is 10-200 on a free plan. I am seeing some failures in the calculation after merging the latest change, although the same change was working last night :/ lowering it to 200 here to minimise flake.

Related to ENG-414


jeanregisser (approved):
🚀

PR #222, opened 2025-06-05
feat: use Redis to cache block number

kathaypacific:
This PR allows the fetchReferrals and calculateKpi scripts to take optional redis configs, and applies a redis cache to the DefiLlama requests. This means that across calculations, we will no longer need to hit DefiLlama to get the block number if it was computed in a previous iteration. This should reduce computation and speed up the calculation.

Related to ENG-414


scripts/calculateKpi.ts
```diff
@@ -136,6 +151,18 @@ async function getArgs() {
     .option('datadir', {
       description: 'Directory to save data',
       default: 'rewards',
+    })
+    .option('use-redis', {
+      boolean: true,
```
bakoushin, line 156: setting this to `true` by default will require Redis instance to be started on localhost -- otherwise the script will fail (i guess). is that what we want?

bakoushin (approved):
🏎️

scripts/calculateKpi.ts
kathaypacific: i also thought this but it seems this configuration means that the presence of the config means "true" (i.e. you can pass `--use-redis`, rather than `--use-redis=true`). i've tested with omitting this argument in the script and redis is then not enabled

bakoushin: ah, sure. I didn't noticed it's `boolean` and not `default` 🙈

scripts/calculateKpi.ts
```diff
@@ -136,6 +151,18 @@ async function getArgs() {
...
+      implies: ['redis-port', 'redis-host'],
+    })
+    .option('redis-port', {
+      type: 'number',
+      default: 6379,
+    })
+    .option('redis-host', {
+      type: 'string',
+      default: '127.0.0.1',
```
jeanregisser, line 165: nit: I would recommend to use `--redis-connection redis://127.0.0.1:6379` string instead.
1 single argument to indicate the same info.

jeanregisser (approved):
Maybe dumb question, since we have our DivviRegistry indexer, do you think we could skip this?
And query it for the needed timestamp or block number of when the referral happened?
Which I think is the main cause of requests to DefiLlama since it depends on the user.

Though it's likely the indexer has limits we don't know about yet 😅

scripts/calculateKpi.ts
kathaypacific: oh yeah good idea! i'll update

kathaypacific:
> Maybe dumb question, since we have our DivviRegistry indexer, do you think we could skip this? And query it for the needed timestamp or block number of when the referral happened? Which I think is the main cause of requests to DefiLlama since it depends on the user.
> 
> Though it's likely the indexer API has limits we don't know about yet 😅

my understanding is that we still want to fetch the blocks given a timestamp for the network(s) that the kpi is being calculated on...so we'd need to deploy an indexer per supported network? it's probably a useful thing to have (allows us to remove `submitReferral` etc), but for this particular job i think redis is probably easier...

jeanregisser:
> > Maybe dumb question, since we have our DivviRegistry indexer, do you think we could skip this? And query it for the needed timestamp or block number of when the referral happened? Which I think is the main cause of requests to DefiLlama since it depends on the user.
> > Though it's likely the indexer API has limits we don't know about yet 😅
> 
> my understanding is that we still want to fetch the blocks given a timestamp for the network(s) that the kpi is being calculated on...so we'd need to deploy an indexer per supported network? it's probably a useful thing to have (allows us to remove `submitReferral` etc), but for this particular job i think redis is probably easier...

I meant we'd keep using DefiLlama for requests that are not user specific.
And the DivviRegistry indexer for the user specific request related to the referral timestamp.
Not sure which one is really simpler, trading redis for scripts changes and using the indexer.

PR #225, opened 2025-06-05
feat: Add segmentedKpi field and add to scout game reward output

finnian0826:
`segmentedKpi` field added to KPI output which is a map from string to numbers, so any wanted breakdown of KPI can be included. Then the specific rewards processing file for the campaign can be updated to process this and include it in the rewards.json output to be used in analysis/dashboard. KPI breakdown by blockchain added for scout-game campaign.

**Testing**
Ran locally and rewards.json output was:
```
[
  {
    "referrerId": "0x5e23d5be257d9140d4c5b12654111a4d4e18d9b2",
    "kpi": "12",
    "rewardAmount": "174846083434101646003",
    "baseKpi": "0",
    "celoKpi": "12",
    "polygonKpi": "0"
  },
  {
    "referrerId": "0x124d8fad33e0b9fe9f3e1e90d0fc0055abe8ca8d",
    "kpi": "1",
    "rewardAmount": "50473716668715175624",
    "baseKpi": "0",
    "celoKpi": "1",
    "polygonKpi": "0"
  },
  {
    "referrerId": "0xe6e6c0ddf053822bbbc1f39218f30ea25fca58dd",
    "kpi": "50948",
    "rewardAmount": "11392757558913212937413",
    "baseKpi": "45973",
    "celoKpi": "4568",
    "polygonKpi": "407"
  },
  ...
]
```


scripts/calculateKpi/protocols/aerodrome/index.ts
```diff
@@ -60,13 +60,14 @@ export async function calculateKpi({
   address: string
   startTimestamp: Date
   endTimestampExclusive: Date
-}): Promise<number> {
-  return calculateRevenueDrome({
+}): Promise<{ kpi: number }> {
```
silasbw, line 63: Make a `KpiResult` interface type?

src/resultDirectory.ts
```diff
@@ -98,7 +104,7 @@ export class ResultDirectory {
   }
 
   async readKpi() {
-    return (await this._readCsv(this.kpiFileSuffix)) as KpiRow[]
+    return (await this._readJson(this.kpiFileSuffix)) as KpiRow[]
```
finnian0826, line 106: the nested JSON for segmentedKpi didn't work well with how the csv was stringifying and saving, so thought it would be easier to just read the json in instead.

scripts/types.ts
```diff
@@ -48,12 +48,17 @@ export interface TokenPriceData {
   priceFetchedAt: number
 }
 
+export interface KpiResult {
+  kpi: number
+  segmentedKpi?: { [key: string]: number }
+}
```
jeanregisser, line 54:
```suggestion
/**
 * Represents the result of a KPI computation.
 *
 * @template T - The allowed string keys for the breakdown.
 * - `value`: overall KPI value.
 * - `breakdown`: an optional map from typed segment keys to their KPI values.
 */
export interface KpiResult<T extends string = string> {
  value: number
  breakdown?: Record<T, number>
}
```

This gives us:
- Type safety for known segments per KPI function
- Cleaner dev experience via autocomplete and key validation
- Better documentation
- Flexibility, still works with untyped keys when the generic is omitted

Then we could use it like this:

```ts
type ScoutGameBreakdown = 'base' | 'celo' | 'polygon'

const result: KpiResult<ScoutGameBreakdown> = {
  value: 87,
  breakdown: {
    base: 90,
    celo: 85,
    polygon: 42,
  },
}
```

scripts/types.ts
```diff
@@ -48,12 +48,17 @@ export interface TokenPriceData {
...
+
 export type CalculateKpiFn = (params: {
   address: string
   startTimestamp: Date
   endTimestampExclusive: Date
   redis?: RedisClientType
-}) => Promise<number>
+}) => Promise<KpiResult>
```
jeanregisser, line 61:
```
export type CalculateKpiFn<T extends string = string> = (params: {
  address: string
  startTimestamp: Date
  endTimestampExclusive: Date
  redis?: RedisClientType
}) => Promise<KpiResult<T>>
```

jeanregisser (approved):
Looks great! 🚀 

I'm proposing a typed breakdown for KpiResult to make KPI outputs safer and easier to work with. Should help catch key mismatches early and improve DX. Open to feedback!
And can be done as a follow up PR.

scripts/types.ts
silasbw: oh, good suggestion @jeanregisser :bow:

silasbw (approved):

PR #227, opened 2025-06-06
feat: add redis connection

kathaypacific:
We can't access the google memory store redis from outside the google VPC, however moving this job to GCP would require setting up a new cloud function / memory store / cloud scheduler which just seems like too much work.

This PR adds a redis connection to a redis saas service. The redis connection url is added in GH secrets, I will share the access details with the rest of the team. 

The cost is cheaper than the GCP solution and is way faster to implement. We could eventually decrease our reliance on DefiLlama somehow, but for now adding redis would ensure that we only calculate the referral block for users who were referred since the last CI run. This should speed up the CI a lot.

Related to ENG-414


jeanregisser (approved):
🚀

PR #238, opened 2025-06-10
feat: add lisk campaign to data upload CI

kathaypacific:
Related to ENG-415


scripts/uploadCurrentPeriodKpis.ts
```diff
@@ -24,23 +25,6 @@ interface Campaign {
 }
 
 const campaigns: Campaign[] = [
-  {
```
kathaypacific, line 27: unrelated but this campaign is over and being ignored anyway

PR #239, opened 2025-06-10
feat: add protocol to data upload workflow dispatch

kathaypacific:
The kpi calculation can take a long time for popular campaigns, this is a good problem to have but it is annoying to have to run this locally if you are the reward distribution engineer. This PR adds a `protocol` input to the workflow dispatch, so that we can manually trigger the CI to run the calculation and download the files later from google cloud storage.


jeanregisser (approved):
🚀

PR #247, opened 2025-06-16
feat(DivviRegistry): add off-chain referral support

jeanregisser:
### 🎯 Summary

This PR adds support for **off-chain referral registration** to the `DivviRegistry`, allowing builders to submit referrals using a user-signed message instead of requiring an on-chain transaction. This enables lower-friction referral paths while preserving transparency and auditability. For example, supporting use cases like **Fonbnk cash-in**, where users may not be able to submit a referral directly on-chain.

### ✅ Key Changes

* Allows referrals to be registered using a **user-signed message**, verified **off-chain** by the trusted backend.
* The **signed message is stored on-chain** via calldata, ensuring full transparency without emitting it in logs.
* Reuses existing referral event format (no changes to structure), ensuring **full compatibility** with current Divvi dapp and Dune dashboards.
  * Off-chain referrals are identified by `txHash = 0`. Note: `chainId` is still set so EIP-1271 verification can be done externally too on the right chain.

### 🔒 Security Notes

* The backend is trusted to validate the referral tag and signature before submission.
* This matches the trust model of our existing TX-based path, but now includes public calldata for external verification.

### 🧩 Compatibility

* Fully backward compatible with all current logic and consumers.
* Builders can adopt this off-chain flow without breaking any existing integrations.


contracts/DivviRegistry.sol
```diff
@@ -244,19 +299,127 @@ contract DivviRegistry is
...
+      // Invalid message type for signature verification
+      return false;
+    }
+
+    // Check if user is a contract (potential smart wallet)
+    if (user.code.length > 0) {
+      // Try EIP-1271 verification for smart contracts
+      try
+        IERC1271(user).isValidSignature(messageHash, offchainMessage.signature)
+      returns (bytes4 magicValue) {
```
jophish, line 384: Just for my own knowledge, what's the deal with this `EIP1271_MAGIC_VALUE` const? I see [this here](https://docs.uniswap.org/contracts/v3/reference/periphery/interfaces/external/IERC1271), but I guess I'm a bit confused as to _why_ this function has this sort of interface.

contracts/DivviRegistry.sol
```diff
@@ -244,19 +299,127 @@ contract DivviRegistry is
...
+    }
+  }
+
+  /**
+   * @notice Verify a signature for referral consent
+   * @param user The address that should have signed (EOA or EIP-1271 contract)
+   * @param offchainMessage The off-chain message data containing message, signature, and type
+   * @return valid Whether the signature is valid
+   */
+  function _verifyOffchainReferralSignature(
```
jophish, line 363: So, to my understanding, this is only checking that the message was indeed signed by the user the registration is being made for. And indeed, looking at how `_registerReferral` was modified, this is the extent of the checking that we do. It seems to me that this doesn't actually provide super meaningful on-chain validation, no? As long as we have access to _any_ message/signature from some user, we can still refer them in any arbitrary way. e.g., even if we register referrals honestly for a while, we can always go look at the calldata of previous `batchRegisterReferral` calls, grab messages/signatures from there, and re-call `batchRegisterReferral` with whatever provider/consumer pairs we want by replaying the original messages/signatures. It feels like we should be doing tag validation on-chain (i.e., pulling providers/consumers from within the tag itself), otherwise the security benefits of this seem a little bit superficial. I see that you mention this in the PR description, but wanted to call this out explicitly since it seems kind of important imo.

kathaypacific (approved):
nice!! LGTM! 🕺🏻

contracts/DivviRegistry.sol
jophish: thanks for the great feedback! yeah this all makes sense to me. +1 to eventually minimizing the scope of the relayer to essentially just "subsidizing gas", but at this early stage it does indeed make sense to just punt on full on-chain validation, especially since the current implementation here already reaches "verifiability parity" with out current TX-based flow.

jophish (approved):

kathaypacific (approved):
LGTM, bummer about the cross chain signatures but i like how removing the onchain signature verification makes the contract more straightforward and makes the onchain and offchain paths more consistent with each other

PR #255, opened 2025-06-18
chore: add slack notification on kpi upload CI failure

kathaypacific:
This PR adds a slack notification to the on-call slack channel when the periodic kpi calculation and upload CI fails. I've been keeping an eye on this workflow for the last few weeks and it is quite stable, so I don't expect this alert to fire often.

I followed the integration steps [here](https://tools.slack.dev/slack-github-action/sending-techniques/sending-data-slack-incoming-webhook/post-inline-text-message). Example of slack alert is [here](https://valora-app.slack.com/archives/C02N3AR2P2S/p1750245850795729).

Fixes ENG-392

### DIFF
--- a/.github/workflows/calculate-and-upload-KPIs.yaml
+++ b/.github/workflows/calculate-and-upload-KPIs.yaml
@@ -10,7 +10,6 @@ on:
         required: true
       protocol:
         description: 'Protocol to calculate KPIs for, e.g. celo-pg, scout-game-v0, lisk-v0'
-  pull_request:
 
 # Cancel same in progress run(s) of the workflow to avoid uploading outdated KPIs
 concurrency:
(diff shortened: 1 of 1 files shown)


.github/workflows/calculate-and-upload-KPIs.yaml
```diff
@@ -37,13 +37,11 @@ jobs:
...
-      #   env:
-      #     SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+      - name: Notify Slack on Failure
+        if: failure()
+        uses: slackapi/slack-github-action@v2.1.0
+        with:
+          webhook: ${{ secrets.SLACK_ON_CALL_WEBHOOK_URL }}
+          webhook-type: incoming-webhook
+          payload: |
+            text: "⚠️ <!subteam^S0277QUM4KB> The periodic KPI calculation and upload workflow failed! Please check the run details: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
```
bakoushin, line 47: do we have any operations instructions on what to do in case of failure? e.g. re-run the script, or wait until the next trigger, etc. it would be handy for an engineer not directly involved in this particular project

kathaypacific: i think this is a good callout, i updated the messaging to indicate that we should only investigate when there are persistent failures. i don't anticipate any obvious reason for there to be persistent errors though, and have left it at reviewing the details of the workflow in case the workflow logic changes and we forget to update this message.

PR #265, opened 2025-06-24
chore: first pass a ethcc reward logic

silasbw:


scripts/calculateRewards/divviIntegrationV1.ts
```diff
@@ -0,0 +1,189 @@
...
+const DIVVI_REWARD_POOL_ADDRESS = '0xEd5527Cac28C4CEe3dab472b1f9F80D30Cf0D277'
+
+// TODO(sbw): hardcoded for 10 EURC for now. We should look at the RewardPool,
+// look at the reward token, and get the correct decimals.
+const DIVVI_REWARD_AMOUNT = parseUnits('10', 6)
+
+async function getArgs() {
+  const argv = await yargs
+    .env('')
+    .option('write', {
```
jeanregisser, line 41: nit:
```suggestion
    .option('dry-run', {
```

scripts/calculateRewards/divviIntegrationV1.ts
```diff
@@ -0,0 +1,189 @@
...
+    },
+  }
+
+  await paginateQuery(client, queryForIntegrators, async (response) => {
+    for (const transaction of response.data.logs) {
+      const address = sliceHex(
+        transaction.topics[3]?.toLowerCase() as Address,
+        -20,
+      )
+      referralConsumers.add(address)
```
jeanregisser, line 90: nit: I think the address wouldn't be 0x prefixed this way. We could decode the log instead.

scripts/calculateRewards/divviIntegrationV1.ts
```diff
@@ -0,0 +1,189 @@
...
+      chain: base,
+      transport: http(),
+    })
+    const txHash = await walletClient.writeContract({
+      address: DIVVI_REWARD_POOL_ADDRESS,
+      abi: rewardPoolAbi,
+      functionName: 'addRewards',
+      args: [referralConsumersThatNeedRewards, rewardAmounts, [0n]],
+    })
+    console.log('writeContract successful', txHash)
```
jeanregisser, line 169: We need to wait for the txHash confirmation.

jeanregisser (commented):
Awesome! 🚀 
'll work toward wrapping this (CF or similar) so it can be invoked automatically.

jeanregisser (approved):

PR #266, opened 2025-06-25
refactor: move builder allow list to reward calc step

kathaypacific:
The rationale behind this change is detailed in ENG-471 so I'll not repeat myself but essentially this change is required so that the upcoming Celo PG reward distribution can be less time consuming by usingthe uploaded kpi calculations.

Fixes ENG-471

### DIFF
--- a/.gcloudignore
+++ b/.gcloudignore
@@ -1,0 +1,18 @@
+# This file specifies files that are *not* uploaded to Google Cloud
+# using gcloud. It follows the same syntax as .gitignore, with the addition of
+# "#!include" directives (which insert the entries of the given .gitignore-style
+# file at that point).
+#
+# For more information, run:
+#   $ gcloud topic gcloudignore
+#
+.gcloudignore
+# If you would like to upload your .git directory, .gitignore file or files
+# from your .gitignore file, remove the corresponding line
+# below:
+.git
+.gitignore
+
+node_modules
+#!include:.gitignore
+!dist
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -104,3 +104,33 @@ jobs:
       #     path: coverage/lcov-report
       # - name: 'Upload coverage to Codecov'
       #   uses: codecov/codecov-action@v3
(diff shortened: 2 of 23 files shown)


scripts/calculateRewards/celoPG.ts
```diff
@@ -102,22 +102,10 @@ function parseArgs() {
...
-          .flatMap((s) => s.split(',').map((item) => item.trim()))
-          .filter(Boolean)
-      },
-    })
-    .option('fail-on-exclude', {
-      description:
-        'Fail if any of the excluded addresses are found in the referral events',
-      type: 'boolean',
-      default: false,
+    .option('builder-allowlist-file', {
```
kathaypacific, line 105: i figure that if there is an inclusion list from karma gap, then we don't really need our own exclusion list on top of that.

scripts/calculateRewards/celoPG.ts
```diff
@@ -152,19 +139,18 @@ export async function main(args: ReturnType<typeof parseArgs>) {
...
-    }).map(({ referrerId }: { referrerId: string }) =>
-      referrerId.toLowerCase(),
-    ),
-  ) as string[]
+  const allowList = args.builderAllowListFile
+    ? (parse(readFileSync(args.builderAllowListFile, 'utf-8').toString(), {
+        skip_empty_lines: true,
+        columns: true,
+      }).map(
+        ({ referrerId }: { referrerId: Address }) => referrerId,
```
finnian0826, line 147:
```suggestion
        ({ referrerId }: { referrerId: Address }) => referrerId.toLowerCase(),
```
?

finnian0826 (approved):
lgtm! 

@silasbw  and I were talking about optimimizing the reward distribution by having the periodic job save the safe-transactions.json so that could be used directly for distribution, but it seems like with the need for an allow/exclude list not used in the periodic job we can't do this?

Also tangentially related, should the scout-game reward calculation be using the exclude list to not include bad addresses in rewards? Currently it looks like it doesn't have any filtering.

PR #267, opened 2025-06-25
test: add unit tests for upload kpi script

kathaypacific:
This adds unit tests to validate the core logic of the uploadCurrentPeriodKpis script. The goal is to ensure the workflow behaves as expected and to encode those expectations directly in code. While this workflow was originally built for analytics (a lower-stakes use case), we’re now considering using it for reward distributions, so it's worth adding additional guardrails to ensure correctness.


scripts/uploadCurrentPeriodKpis.test.ts
```diff
@@ -0,0 +1,274 @@
...
+    it('should skip campaigns that are not active', async () => {
+      await uploadCurrentPeriodKpis(
+        {
+          ...defaultArgs,
+          calculationTimestamp: '2025-07-10T12:00:00Z', // Timestamp after scout-game-v0 campaign ends
+        },
+        mockCampaigns,
+      )
+
+      // Should not call fetchReferrals for any campaigns since they're not active
```
finnian0826, line 249: Isn't celo-pg active and that is why it is being called once?

finnian0826: Could use 2025-08-10 and then check that it neither of the below are called.

finnian0826 (approved):
Approved with one questions and one note/fix about a test

scripts/uploadCurrentPeriodKpis.test.ts
kathaypacific: i updated this comment, it was outdated!

PR #300, opened 2025-07-09
chore: calculate kpi manual dispatch CI accepts multiple protocols

kathaypacific:
As discussed in retro yesterday


jeanregisser (approved):
🚀

PR #305, opened 2025-07-11
feat: add kpi calc for tether campaign

kathaypacific:
As the title.

Related to ENG-495

### DIFF
--- a/docs/calculate-kpi/README.md
+++ b/docs/calculate-kpi/README.md
@@ -18,4 +18,5 @@
 - [rhino](rhino.md)
 - [scoutGameV0](scoutGameV0.md)
 - [somm](somm.md)
+- [tetherV0](tetherV0.md)
 - [velodrome](velodrome.md)
--- a/docs/calculate-kpi/tetherV0.md
+++ b/docs/calculate-kpi/tetherV0.md
@@ -1,0 +1,93 @@
+[**Divvi Protocol - KPI calculation functions**](README.md)
+
+---
+
+[Divvi Protocol - KPI calculation functions](README.md) / tetherV0
+
+# tetherV0
+
+## Functions
+
+### calculateKpi()
+
+```ts
+function calculateKpi(params): Promise<KpiResult<string>>
+```
+
+Defined in: [tetherV0/index.ts:155](https://github.com/divvi-xyz/divvi-protocol-v0/blob/main/scripts/calculateKpi/protocols/tetherV0/index.ts#L155)
+
+Calculates eligible transaction count for Tether (USDT) activity across multiple networks.
+
+**KPI Unit**: Transaction count (number of eligible transactions) where the net transfer value is >= 1 USDT or USDT0
+
+**Business Purpose**: Measures the volume of significant Tether (USDT) transactions to or from a specific user
+across multiple blockchain networks. This metric quantifies user engagement with the Tether ecosystem and
+supports analysis of stablecoin usage patterns and cross-chain activity.
+
+**Protocol Context**: Tether V0 tracks transaction volume to measure user participation in the stablecoin
+ecosystem across various networks. Transaction counts serve as a proxy for user engagement and economic
+activity, supporting stablecoin adoption analysis and cross-chain usage patterns.
+
+**Networks**: Ethereum Mainnet, Avalanche Mainnet, Celo Mainnet, Unichain Mainnet, Ink Mainnet,
+Optimism Mainnet, Arbitrum One, Berachain Mainnet
+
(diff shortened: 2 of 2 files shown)


scripts/calculateKpi/protocols/tetherV0/index.ts
```diff
@@ -0,0 +1,149 @@
...
+          transactionHashesCounted.add(transactionHash)
+        }
+      }
+    }
+  })
+
+  return totalTransactions
+}
+
+export async function calculateKpi({
```
jeanregisser, line 108: nit: we should document this function like we did for the other campaigns (KPI unit, business purpose, calculation method, etc).

scripts/calculateKpi/protocols/tetherV0/index.ts
```diff
@@ -0,0 +1,149 @@
...
+  // Since we are using events to count transfers, we need to ensure that transactions that emit multiple events are only counted once
+  const transactionHashesCounted = new Set<string>()
+  let totalTransactions = 0
+
+  const query = {
+    transactions: [{ from: [user] }],
+    logs: [
+      {
+        contractAddress: tokenAddress,
+        topics: [[transferEventSigHash], [pad(user, { size: 32 })], [], []],
```
jeanregisser, line 61:
```suggestion
        contractAddress: tokenAddress,
        // transfer from user
        topics: [[transferEventSigHash], [pad(user, { size: 32 })], [], []],
```

scripts/calculateKpi/protocols/tetherV0/index.ts
```diff
@@ -0,0 +1,149 @@
...
+      },
+      {
+        contractAddress: tokenAddress,
+        topics: [[transferEventSigHash], [], [pad(user, { size: 32 })], []],
```
jeanregisser, line 65:
```suggestion
        contractAddress: tokenAddress,
        // transfer to user
        topics: [[transferEventSigHash], [], [pad(user, { size: 32 })], []],
```

scripts/calculateKpi/protocols/tetherV0/index.ts
```diff
@@ -0,0 +1,149 @@
...
+      if (data && transactionHash) {
+        const decodedLog = decodeEventLog({
+          abi: erc20Abi,
+          data: data as Hex,
+          topics: topics as [],
+        })
+        const transferValue = decodedLog.args.value
+
+        if (
+          BigNumber(transferValue).gt(MIN_ELIGIBLE_VALUE_IN_SMALLEST_UNIT) &&
```
jeanregisser, line 95: right?
```suggestion
          BigNumber(transferValue).gte(MIN_ELIGIBLE_VALUE_IN_SMALLEST_UNIT) &&
```

scripts/calculateKpi/protocols/tetherV0/index.test.ts
```diff
@@ -0,0 +1,265 @@
...
+          {
+            data: encodedValueAboveThreshold,
+            topics: [
+              transferEventSigHash,
+              pad(testAddress, { size: 32 }),
+              pad('0x4567890123456789012345678901234567890123' as Address, {
+                size: 32,
+              }),
+              '0x0000000000000000000000000000000000000000000000000000000000000000',
+            ],
```
jeanregisser, line 128: supernit: we could maybe consider using `encodeEventTopics` and `encodeAbiParameters` from viem to make it more readable?

Something similar to what we did in https://github.com/divvi-xyz/divvi-relayer/blob/a55f8fc95a8b16547336e22207e4af0f23b324a2/src/services/divviRegistry.test.ts#L98-L113

scripts/calculateKpi/protocols/tetherV0/index.test.ts
```diff
@@ -0,0 +1,265 @@
...
+            transactionHash: '0xabc123', // Same transaction hash
+          },
+        ])
+        await onPage(mockResponse)
+      })
+
+      const result = await calculateKpi(defaultProps)
+
+      // Should count as 1 transaction per network, not 2
```
jeanregisser, line 197: Edge casey: should we calculate the net value transfer for the user?
And accept a transaction that come with more than 1 transfer to reach the minimum value?

jeanregisser (approved):
Awesome! 🚀

scripts/calculateKpi/protocols/tetherV0/index.test.ts
kathaypacific: this is a great point, i've updated the logic to count net transfers per tx!

PR #312, opened 2025-07-14
feat: add function to get rewards exclude list

kathaypacific:
This PR facilitates Valora / VEarn's integration with Divvi by making it ineligible for rewards:
- turn the ofac sanctioned addresses script into a function that returns a list of addresses that are excluded from divvi rewards. currently, this should be OFAC sanctioned addresses + the valora entities.
- there are some addresses that we never want to reward, e.g. ofac sanctioned addresses and for now, valora. this will be solved by a campaign specific inclusion list (e.g. Karma Gap) but when the campaign has none, we should use this function and write the results to the rewards directory. in this PR, the celo-pg and lisk campaigns have been updated to use this function.
- remove the `failOnExclude` behaviour and replace with logging, to remove manual actions that block automating the reward distribution process.

The Dune dashboard displays campaign data using the rewards.json, which means that we should _not_ filter kpi data for excluded referrers. Instead, we should only discount the kpi data in the final reward calculation.

Fixes ENG-504


jeanregisser (approved):
LGTM 🚀

bakoushin:
side note: I recall the fail-on-exclude was discussed as a way to draw attention to some "unwanted" participation, like the one from the OFAC list, because we may want to investigate the situation before distributing rewards.

which makes me think that not all exclusion lists are equal. 

probably we should go inverse: always fail on ofac entries, and have a flag to disable that.

kathaypacific:
> side note: I recall the fail-on-exclude was discussed as a way to draw attention to some "unwanted" participation, like the one from the OFAC list, because we may want to investigate the situation before distributing rewards.
> 
> which makes me think that not all exclusion lists are equal.
> 
> probably we should go inverse: always fail on ofac entries, and have a flag to disable that.

this is an interesting side note...i read this yesterday and have been thinking about it, but i still don't understand what we'd "investigate" if ofac sanctioned addresses did participate. it's not like we can remove their participation if they've added the divvi referral tag to transactions already. if we are not rewarding them by default as this PR does, then is there any point in an investigation? 

i suppose a broader question is whether we want to allow ofac sanctioned addresses to participate at all, but it seems like we should build guardrails for that separate to the reward step (probably earlier in the funnel, in the register entity step).

one thing we definitely want to do is automate parts of the reward distribution more. with this change, if there is no explicit requirement for an inclusion list for the campaign, then we can directly use the safe transactions on the CI (rather than requiring someone to manually run the reward calculation locally, in order to be able to fail on ofac addresses, in order to do an investigation)

PR #313, opened 2025-07-14
feat: Add kpi/reward calculation for blaze the base campaign

finnian0826:
Mostly copied over from celo-pg

Updated to match the campaigns params [here](https://app.divvi.xyz/campaign-overview?providerAddress=0xce56ed47c8f2ee8714087c9e48924b1a30bc455c):
- 1000 USDC in one distribution
- square root reward calculation


scripts/calculateKpi/protocols/blaze-the-base/index.ts
```diff
@@ -0,0 +1,31 @@
+import { RedisClientType } from '@redis/client'
+import { KpiResult, NetworkId } from '../../../types'
+import { getBlockRange } from '../utils/events'
+import { fetchNetworkMetrics } from '../utils/networks'
+
+export async function calculateKpi({
```
jeanregisser, line 6: nit: we should document this function like we did for the other campaigns (KPI unit, business purpose, calculation method, etc).

This affects the public docs we publish about this: https://docs.divvi.xyz/protocol/calculate-kpi/

PR #318, opened 2025-07-18
feat: upload safe txs in CI

kathaypacific:
This change uploads the generated safe-transactions.json file during the kpi calculation CI workflow. For campaigns that do not require an inclusion list at the time of reward calculation (i.e. all current campaigns), this change makes it so that you can download the file from the CI directly without having to run any scripts locally. One less step for reward distributions :)


jeanregisser (approved):
🚀

PR #321, opened 2025-07-18
refactor: calculate kpi returns kpi per builder per user

kathaypacific:
This PR changes the calculate kpi functions to return a map of kpis per referrer (per user) rather than a singular kpi value.

Related to ENG-518.


kathaypacific:
Closing in favour of #327

PR #338, opened 2025-07-22
chore: add get referrer from tx logic for tether campaign

kathaypacific:
This PR adds the logic for getting the referrer from the transaction. I took the code from the `submitReferral` endpoint and tried to preserve as much as possible while removing the offchain portion and support for tag format 0. for simplicity, the Tether campaign will expect builders to use v2 of the referral sdk. We can add these files to a shared module eventually.

To make the review easier, here are the files that are added:
- `alignMagicPrefixAndPadHex.ts` - copied from [here](https://github.com/divvi-xyz/divvi-relayer/blob/main/src/endpoints/submitReferral/alignMagicPrefixAndPadHex.ts) with no changes
- `getTransactionInfo.ts` - copied from [here](https://github.com/divvi-xyz/divvi-relayer/blob/main/src/endpoints/submitReferral/getTransactionInfo.ts) with a change to move `TransactionInfo` inside the file and turn the http errors into standard errors
- `getUserOperations.ts` - copied from [here](https://github.com/divvi-xyz/divvi-relayer/blob/main/src/endpoints/submitReferral/getUserOperations.ts) with no changes
- `parseReferral.ts` - copied from [here](https://github.com/divvi-xyz/divvi-relayer/blob/main/src/endpoints/submitReferral/parseReferral.ts) with changes to remove logic for offchain path and format tag v0 path
- the changes in `getReferrerIdFromTx.ts` mostly follows the logic from [here](https://github.com/divvi-xyz/divvi-relayer/blob/029088c86b39c2b3b5d6032dead08833ce74ed05/src/endpoints/submitReferral/index.ts#L260-L356), minus the support for offchain and tag format v0

Related to ENG-518


scripts/calculateKpi/protocols/tetherV0/alignMagicPrefixAndPadHex.ts
```diff
@@ -0,0 +1,24 @@
+import type { Hex } from 'viem'
+import { DIVVI_MAGIC_PREFIX } from './parseReferral'
```
kathaypacific, line 2: idk if we care about require cycles but it is more obvious that we have one here now 😅

scripts/calculateKpi/protocols/tetherV0/getUserOperations.ts
```diff
@@ -0,0 +1,141 @@
+import {
```
kathaypacific, line 1: this is copied and pasted from https://github.com/divvi-xyz/divvi-relayer/blob/main/src/endpoints/submitReferral/getUserOperations.ts

scripts/calculateKpi/protocols/tetherV0/parseReferralTag/getTransactionInfo.ts
```diff
@@ -0,0 +1,111 @@
+import {
```
kathaypacific, line 1: this is copied from https://github.com/divvi-xyz/divvi-relayer/blob/main/src/endpoints/submitReferral/getTransactionInfo.ts, and for simplicity i also moved the `TransactionInfo` type here (it is in a types.ts file in the relayer)

scripts/calculateKpi/protocols/tetherV0/parseReferralTag/getUserOperations.ts
```diff
@@ -0,0 +1,141 @@
+import {
```
kathaypacific, line 1: this is copied from https://github.com/divvi-xyz/divvi-relayer/blob/main/src/endpoints/submitReferral/getUserOperations.ts

scripts/calculateKpi/protocols/tetherV0/parseReferralTag/parseReferral.ts
```diff
@@ -0,0 +1,226 @@
...
+
+export type ParseReferralResult =
+  | { referral: ParsedReferral; tagFormatId: SupportedFormatId; error?: never }
+  | {
+      referral?: never
+      tagFormatId: SupportedFormatId | undefined
+      error: { type: ParseReferralErrorType; message: string }
+    }
+
+type ReferralType = 'onchain' | 'offchain'
```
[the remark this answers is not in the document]
kathaypacific: i had some internal debates about whether having this dead code is better or worse for readability - on one hand, leaving it reduces the diffs with the relayer code. but on the other hand, reading this code in isolation is way more confusing. i think you're right to expect the offchain code path to be properly removed though, i've updated it

PR #344, opened 2025-07-22
chore: Update hardhat to use etherscan v2 and add mantle

finnian0826:
Slack thread [here](https://valora-app.slack.com/archives/C025V1D6F3J/p1753310287276109)

TLDR;
Etherscan [moved to v2 api](https://docs.etherscan.io/etherscan-v2/v2-quickstart), can no longer get an api key for a specific explorer (and existing ones have been deprecated), need to use the etherscan one for all. Because of that needed to update the hardhat-verify version to use the new API. But this hardhat update breaks the verification for non [etherscan supported](https://docs.etherscan.io/etherscan-v2/supported-chains) chains. The v1 version still works if there is a map of apiKeys, so for vana and morph (and any future v1 chains), use `--config hardhat.v1.config.ts`

This PR also adds mantle support


finnian0826:
> This works but have you tried making two config files? E.g., for morph we can have `hardhat.v1.config.js` and then can we pass the custom file name to hardhat's `--config` option.

Ahh I didn't realize there was the --config option, updated to do that instead

PR #354, opened 2025-07-23
feat: extend slices script to issue user slices

satish-ravi:
For ENG-550


src/resultDirectory.ts
```diff
@@ -129,11 +133,15 @@ export class ResultDirectory {
     return copyFile(fileName, this.includeListFilePath(fileName))
   }
 
-  async writeBuilderSlices(slices: any[]) {
-    await mkdir(dirname(this.rewardsFileSuffix), { recursive: true })
+  async writeSlices(slices: any[], type: 'builder' | 'user') {
```
silasbw, line 136: don't feel strongly, but the "pattern" in the file/class is to be overly verbose.....so there would be `writeBuilderSlices` and `writeUserSlices`.  i realize that probably makes the pattern you'r using for calling more annoying (because you can't just pass "builder" or "user" through).

silasbw (approved):

PR #358, opened 2025-07-25
chore: add total transactions to lisk kpi metadata

kathaypacific:
This logic is copied from the celo pg campaign. it is needed so that totalTransactions appears in the reward.json and can be used in Dune to display total transactions for the campaign.

Related to ENG-557


jeanregisser (approved):
LGTM 🚀

PR #365, opened 2025-07-28
feat: Make generic reward functions for periodic script to use

finnian0826:


scripts/uploadCurrentPeriodKpis.ts
```diff
@@ -9,23 +9,24 @@ import yargs from 'yargs'
 import { BigNumber } from 'bignumber.js'
 import { ResultDirectory } from '../src/resultDirectory'
 import { main as calculateRewardsCeloPG } from './calculateRewards/celoPG'
-import { main as calculateRewardsScoutGame } from './calculateRewards/scoutGameV0'
```
finnian0826, line 12: Should the now unused reward calculation files be removed? Or is there any reason to keep them?

scripts/calculateRewards/linearProportionalRewards.ts
```diff
@@ -0,0 +1,57 @@
+import { BigNumber } from 'bignumber.js'
```
jophish, line 1: This file and the following one are identical besides line 25, no? At the risk of overabstracting, can we consolidate the two files? It seems like the pattern laid out in these two files is basically what we'll want to use regardless of the actual contents of the reward computation we plug in on line 25.

scripts/uploadCurrentPeriodKpis.ts
jophish: I would say scrap them since they add a lot of clutter (and this repo already has enough of that 😅) but curious if anyone else has other opinions.

scripts/uploadCurrentPeriodKpis.ts
```diff
@@ -42,6 +43,8 @@ const campaigns: Campaign[] = [
       {
         startTimestamp: '2025-05-15T00:00:00Z',
         endTimestampExclusive: '2025-06-01T00:00:00Z',
+        rewardPoolAddress: '0xc273fB49C5c291F7C697D0FcEf8ce34E985008F3',
+        rewardAmountInWei: '25000000000000000000000',
```
jophish, line 47: To make this a bit less error prone (and readable) can we use something like `parseEther(25000)` here and elsewhere in this file?

scripts/calculateRewards/linearProportionalRewards.ts
finnian0826: ahh yes I did this in the following one but not here, just consolidated them to one!

scripts/uploadCurrentPeriodKpis.ts
finnian0826: added this for all that have 18 decimals, and added comments explaining the others (like tether campaign where it is USDT with 6 decimals, so couldn't use parseEther

scripts/uploadCurrentPeriodKpis.ts
finnian0826: I think will maybe create a cleanup follow-up ticket, and see if anyone has reasoning to keep stuff around. We also need to clean up the unused kpi stuff, and protocolFilters are essentially unused so maybe that can be cleaned up to

scripts/uploadCurrentPeriodKpis.ts
satish-ravi: you can use `parseUnits` for USDT, so that way the decimals are also explicit. E.g., `parseUnits('25000', 6)`

scripts/uploadCurrentPeriodKpis.ts
```diff
@@ -9,23 +9,24 @@ import yargs from 'yargs'
...
   rewardsPeriods: {
     startTimestamp: string
     endTimestampExclusive: string
+    rewardPoolAddress: string
+    rewardAmountInWei: string
     calculateRewards?: (args: {
       resultDirectory: ResultDirectory
       startTimestamp: string
       endTimestampExclusive: string
+      rewardPoolAddress: string
```
satish-ravi, line 28: do we need reward pool address per reward period?

satish-ravi (approved):

scripts/uploadCurrentPeriodKpis.ts
finnian0826: no, this is fixed in https://github.com/divvi-xyz/divvi-protocol/pull/394/files where this is consolidated with your other campaign file

PR #370, opened 2025-07-29
feat: Check qualifying transactions for referrals

finnianjs:
Tested locally with:
```
yarn ts-node ./scripts/fetchReferrals.ts --protocol base-v0 --start-timestamp 2025-06-30T00:00:00Z --end-timestamp 2025-07-30T00:00:00Z
```

Also ran for celo-pg (no qualifying tx check) to make sure still worked as expected.

In fetchReferrals:

if qualifying tx check:
- get all referral events from contract to get all referred users
- pass these users to the campaign specific qualifying tx check
- will return a list of valid referrals

if no qualifying transaction check, no changes to behavior

-------

Added a generic function to find valid referrals for a given chain, added for morph, mantle and base.

I think one implication of this is that off-chain referrals will not count for these campaigns, since it looks at transactions from the user which don't exist for that case.

### DIFF
--- a/scripts/fetchReferrals.ts
+++ b/scripts/fetchReferrals.ts
@@ -10,8 +10,6 @@ import { closeRedisClient, getRedisClient } from '../src/redis'
 import { findQualifyingNetworkReferral } from './findQualifyingReferral/qualifyingNetworkReferral'
 import { RedisClientType } from '@redis/client'
 
-// @ts-expect-error unused pending optimization
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
 const protocolToQualifyingReferralFinder: Partial<
   Record<
     Protocol,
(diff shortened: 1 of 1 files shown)


scripts/types.ts
```diff
@@ -81,7 +81,7 @@ export interface ReferralEvent {
   userAddress: string
   timestamp: number
   referrerId: string
-  protocol: Protocol
+  protocol?: Protocol
```
finnian0826, line 84: Not used anywhere, makes qualifying tx finding messy so just making it optional

scripts/findQualifyingReferral/qualifyingNetworkReferral.ts
```diff
@@ -0,0 +1,67 @@
...
+        const block = response.data.blocks[i]
+        if (!tx.hash || !block.timestamp) {
+          continue
+        }
+        const referrerId = await getReferrerIdFromTx(
+          tx.hash as Hex,
+          networkId,
+          true,
+        )
+        if (referrerId !== null) {
```
silasbw, line 55: don't we need to check that there isn't already another qualifyingReferal for the userAddress?

silasbw: oh, nvm i see :facepalm:

PR #373, opened 2025-07-30
chore: work towards optimizing qualifyingNetworkReferral

silasbw:
To test:
```
yarn ts-node ./scripts/fetchReferrals.ts --protocol base-v0 --start-timestamp 2025-06-30T00:00:00Z --end-timestamp 2025-07-30T00:00:00Z
```

### DIFF
--- a/scripts/findQualifyingReferral/qualifyingNetworkReferral.ts
+++ b/scripts/findQualifyingReferral/qualifyingNetworkReferral.ts
@@ -1,12 +1,17 @@
-import { BlockField, TransactionField } from '@envio-dev/hypersync-client'
+import {
+  BlockField,
+  TransactionField,
+} from '@envio-dev/hypersync-client'
 import { getBlockRange } from '../calculateKpi/protocols/utils/events'
 import { NetworkId, ReferralEvent } from '../types'
 import { getHyperSyncClient } from '../utils'
 import { paginateQuery } from '../utils/hypersyncPagination'
 import { getReferrerIdFromTx } from '../calculateKpi/protocols/tetherV0/parseReferralTag/getReferrerIdFromTx'
-import { Hex } from 'viem'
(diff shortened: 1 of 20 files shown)


scripts/findQualifyingReferral/qualifyingNetworkReferral.ts
```diff
@@ -32,22 +34,64 @@ async function findQualifyingNetworkReferralForUser({
...
+    fromBlock: startBlock,
+    toBlock: endBlockExclusive,
   }
   await paginateQuery(client, query, async (response) => {
     for (let i = 0; i < response.data.transactions.length; i++) {
       const tx = response.data.transactions[i]
       const block = response.data.blocks[i]
-      if (!tx.hash || !block.timestamp) {
+      if (!tx || !block) {
+        // was seeing weird behavior where response.data.blocks[i] was undefined, the last entry of blocks was missing??
```
satish-ravi, line 51: this is strange, we should probably switch to use `paginateEventsQuery` over `paginateQuery`, which I believe includes block and tx on the same object and maybe this issue doesn't happen.

satish-ravi (approved):

scripts/findQualifyingReferral/qualifyingNetworkReferral.ts
finnian0826: [This](https://github.com/divvi-xyz/divvi-protocol-v0/blob/main/scripts/utils/referrals.ts#L109-L114) makes me think that maybe it happened with `paginateEventsQuery` too, but I can use it instead to hopefully be safer

satish-ravi: Actually blocks and transactions array aren't 1:1, the blocks array only include unique blocks, thats why it has lesser size than transactions array. I have a fix in #391

PR #390, opened 2025-08-04
feat(cloud-functions): Add updateDivviEntities endpoint for updating entity metadata

jh2oman:
https://linear.app/valora/issue/ENG-581/automate-updating-app-names-in-dune

Cloud function that can be called to copy data from the Builder Info Google sheet to https://storage.googleapis.com/divvi-campaign-data-production/kpi/divvi-entities.json which is used in Dune queries

### DIFF
--- a/scripts/calculateKpi/protocols/tetherV0/parseReferralTag/parseReferral.ts
+++ b/scripts/calculateKpi/protocols/tetherV0/parseReferralTag/parseReferral.ts
@@ -1,7 +1,7 @@
 import { Address, bytesToHex, decodeAbiParameters, Hex, hexToBytes } from 'viem'
 import { alignMagicPrefixAndPadHex } from './alignMagicPrefixAndPadHex'
 
-const SUPPORTED_FORMAT_IDS = [1] as const
+const SUPPORTED_FORMAT_IDS = [0, 1] as const
 // Magic prefix is keccak256("divvi").slice(2, 10)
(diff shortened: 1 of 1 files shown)


cloud-functions/endpoints/updateDivviEntities/index.ts
```diff
@@ -0,0 +1,181 @@
...
+  // Get all values from the sheet
+  const sheetsResponse = await sheets.spreadsheets.values.get({
+    spreadsheetId,
+    range,
+  })
+
+  const values = sheetsResponse.data.values || []
+
+  // Skip header row if it exists and map the data to our interface
+  const rows: SheetRow[] = values.slice(1).map((row) => ({
```
finnian0826, line 66: If header row doesn't exist, would this skip the first row?

finnian0826 (approved):
One small question but otherwise looks good.

Did you deploy / test triggering the cloud function? If so if you could add the command used to trigger to the PR description could be nice for in the future if someone else wants to trigger the upload and comes across this PR could just copy the command.

cloud-functions/endpoints/updateDivviEntities/index.ts
jh2oman: it should always skip the first row. I'll adjust the comment

jh2oman:
@finnian0826 I'm going to make a pr for a scheduled build every hour similar to this pr https://github.com/valora-inc/infrastructure/pull/371/files

PR #399, opened 2025-08-07
fix(ofac): turn ofac back on and cache to reduce the number of calls and hopefully avoid 403's

jh2oman:
* turns back on fetching ofac addresses to exclude them from receiving rewards
* cache to reduce the number of calls

https://linear.app/valora/issue/ENG-582/re-enable-ofac-address-fetching

### DIFF
--- a/scripts/utils/divviRewardsExcludedReferrers.ts
+++ b/scripts/utils/divviRewardsExcludedReferrers.ts
@@ -9,9 +9,19 @@ const valoraEntities: { referrerId: Address; shouldWarn?: boolean }[] = []
 const OFAC_SDN_ZIP_URL =
   'https://sanctionslistservice.ofac.treas.gov/api/download/SDN_XML.ZIP'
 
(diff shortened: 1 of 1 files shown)


scripts/utils/divviRewardsExcludedReferrers.ts
```diff
@@ -55,10 +55,7 @@ export async function getDivviRewardsExcludedReferrers(): Promise<
     }
   >
 > {
-  // TODO: The fetch to get OFAC SDN addresses is disabled for now because it
-  // intermittently fails with 403.
-  const ofacSdnAddresses: { referrerId: Address; shouldWarn: boolean }[] = []
-  // const ofacSdnAddresses = await getOfacSdnAddresses()
+  const ofacSdnAddresses = await getOfacSdnAddresses()
```
satish-ravi, line 58: could just cache this in a global var here and avoid all the passing around of args.

satish-ravi (approved):

PR #403, opened 2025-08-07
chore: Update workflow runtime and have tether use period startTimestamp

finnian0826:
tether campaign looks for referrer tag in all txs so no need to start part way through the period, this reduces need for  multiple defi llama calls to get start block which were very slow

### DIFF
--- a/scripts/calculateKpi.ts
+++ b/scripts/calculateKpi.ts
@@ -58,7 +58,9 @@ async function calculateKpiBatch({
         const calculatedKpi = await calculateKpiHandlers[protocol]({
           address: userAddress,
           // if the referral happened after the start of the period, only calculate KPI from the referral block onwards so that we exclude user activity before the referral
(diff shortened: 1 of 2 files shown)


scripts/calculateKpi.ts
```diff
@@ -58,7 +58,9 @@ async function calculateKpiBatch({
         const calculatedKpi = await calculateKpiHandlers[protocol]({
           address: userAddress,
           // if the referral happened after the start of the period, only calculate KPI from the referral block onwards so that we exclude user activity before the referral
+          // except for the tether campaign, since we check for the referral tag on all txs, so start from the start of the period to reduce defi llama calls
           startTimestamp:
+            protocol !== 'tether-v0' &&
```
finnian0826, line 63: this seemed like easiest way without having to change the inputs to the function, but a little janky so I can do a different way if there is a better approach

satish-ravi: this is the only campaign. I guess we can generalize this as we have more in this pattern

satish-ravi (approved):

PR #407, opened 2025-08-08
feat(protocol): Add protocol fee to RewardPool

jophish:
For [ENG-578](https://linear.app/valora/issue/ENG-578/update-rewardpoolfactory-to-set-default-protocol-fee-and-reserve) and [ENG-579](https://linear.app/valora/issue/ENG-579/update-rewardpool-contract-to-collect-fees-when-adding-rewards).

Adds a `protocolFee` and `reserveAddress` to the `RewardPool` contract. When allocating rewards to users, sends a protocol fee to the reserve address _on top_ of the user's reward, such that users always get rewarded the specified amount. The `RewardPoolFactory` contract has been updated to allow us to set the `reserveAddress` and `protocolFee` that are set when deploying a new `RewardPool` contract; these parameters are pulled _from the factory state_, and are not parameters to `createRewardPool` so that end users cannot override them.

Since the `owner` on a `RewardPool` has the ability to update `protocolFee` and `reserveAddress` for a given reward pool, I've updated the `owner` address that was previously passed in to `createRewardPool` on the factory to follow the same pattern as `protocolFee` and `reserveAddress`; it can now be set by the factory admin as a piece of state on the factory, and is passed into the `RewardPool` init method. In a self-service deployment model, this prevents campaign managers from overriding the owner and updating the `protocolFee` or `reserveAddress` on the deployed reward pool.

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "@safe-global/protocol-kit": "^6.1.0",
     "@safe-global/types-kit": "^3.0.0",
     "@valora/http-handler": "^0.0.1",
-    "@valora/logging": "^1.3.18",
+    "@valora/logging": "^1.3.19",
     "axios": "^1.10.0",
     "dotenv": "^17.2.1",
     "googleapis": "^155.0.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -3533,10 +3533,10 @@
   dependencies:
     "@google-cloud/functions-framework" "^3.1.3"
 
-"@valora/logging@^1.3.18":
-  version "1.3.18"
-  resolved "https://registry.yarnpkg.com/@valora/logging/-/logging-1.3.18.tgz#075fbfc7bbf33e63d9c1b92bc1fa96c01daf8f29"
-  integrity sha512-+NUo6ZEeEeeolP+QdPp5OS4evOp6s4b7uMBfwkFRvvAysd++315wfnG5N2Ib+C8zr9r17gc7ryOHcCJFMI13Dg==
+"@valora/logging@^1.3.19":
+  version "1.3.19"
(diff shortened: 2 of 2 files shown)


test/RewardPoolFactory.ts
```diff
@@ -128,50 +137,211 @@ describe(CONTRACT_NAME, function () {
...
-        ),
+        factoryWithStranger.setDefaultProtocolFee(newFee),
+      ).to.be.revertedWithCustomError(
+        factory,
+        'AccessControlUnauthorizedAccount',
+      )
+    })
+
+    it('reverts when non-owner tries to set default reserve address', async function () {
+      const factoryWithStranger = factory.connect(stranger) as typeof factory
```
finnian0826, line 250: should this be done in the beforeEach like it is in RewardPool.ts tests? (could also do for factoryWithOwner)

finnian0826 (approved):
Approving to unblock, but wondering how this will work in practice.

Since we are rewarding the builder with the full value passed to addRewards, and now on top of that taking a fee, how will we ensure that there is enough rewardToken in the pool? The two ways I can think of are:
- the campaign owner needs to "over deposit" into the pool, i.e. deposit enough to cover rewards and fees. So if it is a 10k campaign with 1% fee, needs to deposit 10.1k tokens.
- we take this into account when calculating rewards and calculate with a slightly smaller amount. If 10k campaign reward period with 1% fee, in our calculate rewards script only calculate with ~9.9k to ensure enough left over for fees.

For the first option, the issue seems to be that the pool amount will not match what we say the reward amount is, so the value on the dapp and in our reward script is inconsistent with the contract.

For the second option this seems like just the same as subtracting the fee from the user, so instead if this case we just calculate rewards as normal with the full value, and then subtract out the fee when adding rewards.

Maybe I am missing another option that solves this?

PR #409, opened 2025-08-08
chore: Remove past campaign with no reward periods

finnian0826:
Fixes periodic upload job failure [here](https://github.com/divvi-xyz/divvi-protocol/actions/runs/16840464907/job/47710204669). This campaign was not in the upload script campaign list, so reward periods didn't get moved over to new campaigns file [here](https://github.com/divvi-xyz/divvi-protocol/pull/365/files#diff-e2e59ccdf441770c6a95eff037bbf79e3225a6df2e943e145ce534a3886943c4). The campaign is done so can just be removed.

Also added a check in the periodic script to just skip campaigns with no rewards periods

### DIFF
--- a/src/campaigns.ts
+++ b/src/campaigns.ts
@@ -240,6 +240,14 @@ export const campaigns: Campaign[] = [
     ],
   },
   {
+    providerAddress: '0x5f0a55fad9424ac99429f635dfb9bf20c3360ab8',
+    protocol: 'celo-transactions',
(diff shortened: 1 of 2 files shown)


src/campaigns.ts
```diff
@@ -239,14 +239,6 @@ export const campaigns: Campaign[] = [
       },
     ],
   },
-  {
```
satish-ravi, line 242: with the above check we can leave this in right?

PR #421, opened 2025-08-13
feat(staking): Add BuilderStaking contract

jophish:
For [ENG-589](https://linear.app/valora/issue/ENG-589/create-builderstaking-contract). Adds a `BuilderStaking` contract that allows users to stake $DIVVI on behalf of other users.

### DIFF
--- a/contracts/BuilderStaking.sol
+++ b/contracts/BuilderStaking.sol
@@ -1,0 +1,331 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import {AccessControlDefaultAdminRulesUpgradeable} from '@openzeppelin/contracts-upgradeable/access/extensions/AccessControlDefaultAdminRulesUpgradeable.sol';
+import {UUPSUpgradeable} from '@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol';
+import {Initializable} from '@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol';
+import {SafeERC20} from '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
+import {IERC20} from '@openzeppelin/contracts/token/ERC20/IERC20.sol';
+
+/**
+ * @title Divvi BuilderStaking
+ * @dev Contract for managing $DIVVI token stakes on behalf of builders
+ * @custom:security-contact security@valora.xyz
+ */
+contract BuilderStaking is
+  Initializable,
+  AccessControlDefaultAdminRulesUpgradeable,
+  UUPSUpgradeable
+{
+  using SafeERC20 for IERC20;
+
+  // State variables
+  IERC20 public divviToken;
+  uint256 public stakingThreshold;
+  uint256 public totalStaked;
+
+  // Core mappings - these store the actual stake amounts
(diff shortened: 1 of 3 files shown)


contracts/BuilderStaking.sol
```diff
@@ -0,0 +1,330 @@
...
+  IERC20 public divviToken;
+  uint256 public stakingThreshold;
+  uint256 public totalStaked;
+
+  mapping(address => mapping(address => uint256)) public stakersForBeneficiary; // beneficiary => staker => amount
+  mapping(address => mapping(address => uint256)) public beneficiariesForStaker; // staker => beneficiary => amount
+
+  // Arrays to track relationships for efficient querying
+  mapping(address => address[]) public stakerListForBeneficiary;
+  mapping(address => address[]) public beneficiaryListForStaker;
```
jeanregisser, line 32: How many beneficiaries/stakers before looping over these arrays takes too much gas?
100 is fine, but what about 1,000 and then 10,000?

Should we reconsider storing these or use another data structure?

contracts/BuilderStaking.sol
```diff
@@ -0,0 +1,330 @@
...
+   */
+  function initialize(
+    address _divviToken,
+    address _admin,
+    uint256 _initialThreshold
+  ) external initializer {
+    if (_divviToken == address(0)) revert ZeroAddressNotAllowed();
+    if (_admin == address(0)) revert ZeroAddressNotAllowed();
+
+    __AccessControlDefaultAdminRules_init(0, _admin);
```
jeanregisser, line 73: Should we set the default transfer delay to something higher?

jeanregisser (commented):
Awesome work!

contracts/BuilderStaking.sol
jophish: I was considering this as well. We can of course just rely on indexing to fetch this data, which is imo a more typical approach, but means that the in order to render pertinent information about staking in the UI, we'll need to do that indexer work. That shouldn't be too difficult however, so it seems like we can probably just scrap this bookkeeping and rely on indexing to keep track of this.

contracts/BuilderStaking.sol
jophish: Sure -- can set it to something like one day as a default.

contracts/BuilderStaking.sol
```diff
@@ -0,0 +1,330 @@
...
+
+    // Update stake amounts
+    stakersForBeneficiary[_beneficiary][msg.sender] -= _amount;
+    beneficiariesForStaker[msg.sender][_beneficiary] -= _amount;
+    totalStaked -= _amount;
+
+    // Remove from tracking arrays if stake becomes zero
+    if (beneficiariesForStaker[msg.sender][_beneficiary] == 0) {
+      _removeStakerFromBeneficiary(msg.sender, _beneficiary);
+      _removeBeneficiaryFromStaker(msg.sender, _beneficiary);
```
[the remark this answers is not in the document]
jophish: Ah yeah, that's probably a good idea!

contracts/BuilderStaking.sol
jophish: Actually, I've updated this to be a constructor param, in line with what we do in other contracts.

jeanregisser:
@cursor review

jeanregisser (approved):
🚀

PR #430, opened 2025-08-18
feat: Create linear and sqrt reward libraries for onchain reward calculation

finnianjs:
Adds the first two reward function libraries for on chain reward calculation. 

To use in the reward contract should have the reward pool take a contract address as an input param (the address of the deployed library to use), and use `delegatecall` passing the function name `calculateReward` and the expected input (list of (kpi, referrerAddress) pairs, and a totalRewardAmount). 

Future reward functions added should follow the same pattern of creating a new library and having the same function name and input params. This way we don't need to upgrade/redeploy any of the reward pool stuff, we just deploy the new reward library and use the address in the reward pool.


contracts/mocks/LinearRewardPool.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: MIT
```
finnian0826, line 1: This and the following were added to be able to test the reward libraries, since the libraries need to be used in a contract to test.

finnian0826: Updated now to have a single test contract that takes the reward function address. This can also be reused for future reward function testing by deploying the reward function in the test and then passing the address to this.

contracts/rewardFunctions/IRewardFunction.sol
```diff
@@ -0,0 +1,19 @@
...
+interface IRewardFunction {
+  struct Kpi {
+    uint256 kpi;
+    address referrerAddress;
+  }
+
+  struct Reward {
+    uint256 reward;
+    address referrerAddress;
+  }
```
satish-ravi, line 13: should we consider matching the struct with this https://github.com/divvi-xyz/divvi-protocol/blob/main/contracts/RewardPool.sol#L24-L28? One option is both Kpi and Reward both include idempotency keys but that doesn't seem like something the RewardFunction needs to care about. We could just handle idempotency key in RewardPool contract, but probably extra data processing is required when calling calculateReward. cc @jophish

contracts/rewardFunctions/SqrtReward.sol
```diff
@@ -0,0 +1,46 @@
...
+      });
+    }
+  }
+
+  /**
+   * @dev Calculate the square root of a number
+   * @param x The number to calculate the square root of
+   * @return y The square root of the number
+   */
+  function _sqrt(uint256 x) internal pure returns (uint256 y) {
```
satish-ravi, line 38: I noticed that the RewardPool contract uses Math from OZ. Is that something we should use for sqrt and safe multiplication and divisions (e.g., `mulDiv`)?

libraries/LinearReward.sol
```diff
@@ -0,0 +1,38 @@
...
+    uint256 totalRewardAmount
+  ) external pure returns (Reward[] memory rewards) {
+    rewards = new Reward[](kpis.length);
+    uint256 totalKpi = 0;
+    for (uint256 i = 0; i < kpis.length; i++) {
+      totalKpi += kpis[i].kpi;
+    }
+    for (uint256 i = 0; i < kpis.length; i++) {
+      rewards[i] = Reward({
+        reward: (totalRewardAmount * kpis[i].kpi) / totalKpi,
```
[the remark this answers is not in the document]
satish-ravi: yeah I guess the max dust would be (n - 1, n = num builders). We can probably distribute those with some algorithm, but probably not worth it, especially since the reward amount is in wei

contracts/rewardFunctions/IRewardFunction.sol
jophish: Hmm, I think we should probably do something like have the new `RewardPool` method which accepts KPIs take (kpi, referrerAddress, idempotencyKey) tuples. We can then either pass those full tuples (including idemopotency keys) into the `calculateReward` function, and have it return (reward, referrerAddress, idempotencyKey) tuples, _or_ just pass (kpi, referrerAddress) and have the caller re-attach the idemopotency keys to the rewards. I'm not sure which option is more gas efficient. It might be less work/cleaner to do less data wrangling, however, and just always include the idempotency keys.

contracts/rewardFunctions/LinearReward.sol
```diff
@@ -0,0 +1,32 @@
...
+    uint256 totalKpi = 0;
+    for (uint256 i = 0; i < kpis.length; i++) {
+      totalKpi += kpis[i].kpi;
+    }
+    if (totalKpi == 0) {
+      return rewards;
+    }
+    for (uint256 i = 0; i < kpis.length; i++) {
+      rewards[i] = Reward({
+        reward: (totalRewardAmount * kpis[i].kpi) / totalKpi,
```
jophish, line 27: Can we use [mulDiv](https://docs.openzeppelin.com/contracts/5.x/api/utils#Math-mulDiv-uint256-uint256-uint256-) here?

contracts/rewardFunctions/SqrtReward.sol
jophish: Yeah, we should/could probably use OZ's [sqrt](https://docs.openzeppelin.com/contracts/5.x/api/utils#Math-sqrt-uint256-) method here, as well as `mulDiv` as mentioned above whenever we multiply / divide uint256 values.

libraries/LinearReward.sol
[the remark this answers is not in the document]
jophish: Yeah, as long as we never exceed the `totalRewardAmount`, I think having some leftover dust is fine -- the campaign manager will be able to reclaim it at the end of the campaign, anyhow.

contracts/rewardFunctions/LinearReward.sol
finnian0826: added in [e3b1a6d](https://github.com/divvi-xyz/divvi-protocol/pull/430/commits/e3b1a6ddb7afd33c8b329dd9cb6fbe0db0ae4d4b)

contracts/rewardFunctions/SqrtReward.sol
finnian0826: added in [e3b1a6d](https://github.com/divvi-xyz/divvi-protocol/pull/430/commits/e3b1a6ddb7afd33c8b329dd9cb6fbe0db0ae4d4b)

contracts/rewardFunctions/IRewardFunction.sol
satish-ravi: Thinking more about idempotency keys, I don't think we can have 1 per kpi row, like before had 1 per reward row. When adding rewards, each row was independent, whereas now each kpi row actually affects how everyone receives rewards (and we cant just skip a row). Maybe we just need one idempotency key per reward period?

satish-ravi (approved):

test/SqrtReward.ts
```diff
@@ -0,0 +1,187 @@
...
+      const rewards = await testContract.testCalculateReward(
+        kpis,
+        totalRewardAmount,
+      )
+
+      expect(rewards).to.have.length(3)
+      // total sqrt = 1 + 2 + 3 = 6
+      expect(rewards[0].reward).to.equal(10n) // 60 * 1 / 6
+      expect(rewards[1].reward).to.equal(20n) // 60 * 2 / 6
+      expect(rewards[2].reward).to.equal(30n) // 60 * 3 / 6
```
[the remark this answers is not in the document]
finnian0826: simulating on the existing tether campaign kpi it seems like scaling kpi by 10^6 goes from an average 6% error to 0.003% so seems worth it, I'll add it.

PR #435, opened 2025-08-21
feat(kpi): Batch user hypersync requests in kpi calculation

finnian0826:


scripts/calculateKpi.test.ts
```diff
@@ -28,148 +44,407 @@ describe('_calculateKpiBatch', () => {
...
-        userAddress: '0x456',
-        timestamp: '2024-01-15T00:00:00Z',
-      },
-      {
-        referrerId: 'ref3',
-        userAddress: '0x789',
-        timestamp: '2024-01-15T00:00:00Z',
-      },
-    ]
+  describe('per-user processing', () => {
```
finnian0826, line 47: These are all the same tests as before, just in a describe, the new tests are in the other desrcibe below for the batch processing

scripts/calculateKpi.ts
```diff
@@ -36,59 +39,146 @@ async function calculateKpiBatch({
...
+            startTimestamp,
+            endTimestampExclusive,
+            redis,
+          }),
+        ),
+      )
+
+      results.push(...batchResults.flat())
+    }
+  } else {
```
finnian0826, line 131: default fallback is just the same as what was done before

scripts/calculateKpi/protocols/index.ts
```diff
@@ -12,7 +12,10 @@ import { calculateKpi as calculateKpiCeloTransactions } from './celoTransactions
 import { calculateKpi as calculateKpiRhino } from './rhino'
 import { calculateKpi as calculateKpiScoutGameV0 } from './scoutGameV0'
 import { calculateKpi as calculateKpiLiskV0 } from './liskV0'
-import { calculateKpi as calculateKpiTetherV0 } from './tetherV0'
+import {
+  calculateKpi as calculateKpiTetherV0,
```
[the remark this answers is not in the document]
finnian0826: Do you mean we should only have 1 handler for tether rather than both, or something different?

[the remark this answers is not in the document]
satish-ravi: its just duplicated logic right now right? if a batch handler exists why even have a single user handler?

scripts/calculateKpi/protocols/tetherV0/index.ts
```diff
@@ -338,3 +339,320 @@ export async function calculateKpi({
...
+      }
+      const referral = await getReferrerIdFromTx(
+        transactionHash as Hex,
+        networkId,
+        true,
+        transactionInfo,
+      )
+      if (
+        referral !== null &&
+        users.includes(referral.user.toLowerCase() as Address)
```
satish-ravi, line 629: This doesn't seem right. I believe the intent was to ensure the from address and the referred user was the same and in the existing implementation there's only one user in question. I think this should compare against `from`, but maybe this is redundant since `parseReferral` might guarantee it?

scripts/calculateKpi.test.ts
```diff
@@ -28,148 +42,352 @@ describe('_calculateKpiBatch', () => {
...
+    it('should deduplicate users by address keeping first occurrence', async () => {
+      const eligibleUsers = [
+        {
+          referrerId: 'ref1',
+          userAddress: '0x123',
+          timestamp: '2024-01-10T00:00:00Z', // First occurrence
+        },
+        {
+          referrerId: 'ref2',
+          userAddress: '0x123', // Duplicate address
```
satish-ravi, line 287: this should never happen in practice right? A user can be referred to a campaign only once on the registry contract?

satish-ravi (approved):
Can we also ensure kpi.json is the same (w/ the additional metadata), similar to rewards.json?

scripts/calculateKpi.test.ts
finnian0826: Correct, shouldn't be possible

scripts/calculateKpi/protocols/tetherV0/index.ts
finnian0826: I thought it was to check that the referred user was the same as the user returned from `getReferrerIdFromTx`, which is usually the same as from except for AA transactions. I.e. we are checking that the tx was initiated by a referred user, either the normal way where they are the from address, or parsed from the tx data for AA txs. I think what I have is the batch equivalent of `referral.user === user` in the non-batch function, right?

satish-ravi: if you're comparing against the entire users list, it could be a different referred user being processing in the same batch right?

PR #438, opened 2025-08-22
docs: add more info about fee protocol and its usage

jeanregisser:
Caught up with https://github.com/divvi-xyz/divvi-protocol/pull/407 and was wondering about the intended usage.
Added more docs to make it a bit more clear for the next person.


jophish (approved):
Thanks Jean!!! This is super helpful documentation :)

PR #462, opened 2025-08-29
feat(contracts): add kpi calculation to reward pool

satish-ravi:
RewardPool changes:
- updates the constructor and initializer to take a `rewardFunctionAddress` instead of a `rewardFunctionId` which must be an implementation of `IRewardFunction`
- adds an `addKpis` method which can be called with a set of (referrer, kpi) pairs for a reward period (identified by start and end timestamp)
  - performs an upsert of kpis for the period
  - 0 kpis are treated as no kpis (can be used for deletion)
  - only allows adding kpis if a reward period is not already processed and rewards are issued
- adds a `processRewardPeriod` method which can be called with a reward period (start and end timestamp) and a reward amount
  - calls `IRewardFunction`'s `calculateReward` method by collecting all the kpis for the given period and issues rewards
  - marks the period as processed so it doesn't end up issuing rewards again
- adds a `setRewardFunctionAddress` to allow updating `rewardFunctionAddress`
 
RewardPoolFactory changes:
- updates to match the constructor changes in RewardPool

### DIFF
--- a/contracts/RewardPool.sol
+++ b/contracts/RewardPool.sol
@@ -6,6 +6,7 @@ import {ReentrancyGuard} from '@openzeppelin/contracts/utils/ReentrancyGuard.sol
 import {SafeERC20} from '@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol';
 import {IERC20} from '@openzeppelin/contracts/token/ERC20/IERC20.sol';
 import {Math} from '@openzeppelin/contracts/utils/math/Math.sol';
+import {IRewardFunction} from './rewardFunctions/IRewardFunction.sol';
 
 /**
  * @title Divvi RewardPool
@@ -44,11 +45,14 @@ contract RewardPool is AccessControl, ReentrancyGuard {
   // State variables
   address public poolToken;
   bool public isNativeToken;
-  bytes32 public rewardFunctionId;
+  address public rewardFunctionAddress;
   uint256 public timelock;
   uint256 public totalPendingRewards;
   mapping(address => uint256) public pendingRewards;
   mapping(bytes32 => bool) public processedIdempotencyKeys;
+  mapping(bytes32 => mapping(address => uint256)) public periodKpis;
+  mapping(bytes32 => address[]) private periodUsers;
+  mapping(bytes32 => bool) public processedPeriods;
 
   // Protocol fee state variables
   uint256 public protocolFee;
@@ -57,7 +61,7 @@ contract RewardPool is AccessControl, ReentrancyGuard {
   // Events
   event PoolInitialized(
     address indexed poolToken,
-    bytes32 rewardFunctionId,
+    address rewardFunctionAddress,
     uint256 timelock
   );
   event Deposit(uint256 amount);
@@ -92,6 +96,24 @@ contract RewardPool is AccessControl, ReentrancyGuard {
     uint256 feeAmount,
     uint256 protocolFee
   );
+  event RewardFunctionAddressUpdated(
+    address newRewardFunctionAddress,
+    address previousRewardFunctionAddress
+  );
+  event AddKpi(
+    address indexed user,
+    bytes32 indexed rewardPeriodKey,
+    uint256 amount,
+    uint256 rewardPeriodStart,
+    uint256 rewardPeriodEndExclusive,
+    bytes32 kpiFunctionId
+  );
+  event RewardPeriodProcessed(
+    bytes32 indexed rewardPeriodKey,
+    uint256 rewardPeriodStart,
+    uint256 rewardPeriodEndExclusive,
+    uint256 totalRewardAmount
+  );
 
   // Errors
   error AmountMismatch(uint256 expected, uint256 received);
@@ -120,6 +142,12 @@ contract RewardPool is AccessControl, ReentrancyGuard {
   error AlreadyInitialized();
   error InvalidProtocolFee(uint256 fee);
   error InvalidReserveAddress();
+  error InvalidRewardFunctionAddress();
+  error RewardPeriodInvalid(
+    uint256 rewardPeriodStart,
+    uint256 rewardPeriodEndExclusive
+  );
+  error RewardPeriodAlreadyProcessed(bytes32 rewardPeriodKey);
 
   // This is needed to prevent the implementation from being initialized
   bool private initialized;
@@ -127,7 +155,7 @@ contract RewardPool is AccessControl, ReentrancyGuard {
   /**
    * @dev Initializes the contract
    * @param _poolToken Address of the token used for rewards
-   * @param _rewardFunctionId Bytes32 identifier of the reward function (e.g. git commit hash)
+   * @param _rewardFunctionAddress Implementation address of the reward function
    * @param _owner Address that will have DEFAULT_ADMIN_ROLE
    * @param _manager Address that will have MANAGER_ROLE
    * @param _timelock Timestamp when manager withdrawals will be allowed
@@ -140,7 +168,7 @@ contract RewardPool is AccessControl, ReentrancyGuard {
    */
   function initialize(
     address _poolToken,
-    bytes32 _rewardFunctionId,
+    address _rewardFunctionAddress,
     address _owner,
     address _manager,
     uint256 _timelock,
@@ -156,19 +184,19 @@ contract RewardPool is AccessControl, ReentrancyGuard {
 
     poolToken = _poolToken;
     isNativeToken = (_poolToken == NATIVE_TOKEN_ADDRESS);
-    rewardFunctionId = _rewardFunctionId;
+    rewardFunctionAddress = _rewardFunctionAddress;
 
     _setTimelock(_timelock);
     _setProtocolFee(_protocolFee);
     _setReserveAddress(_reserveAddress);
 
-    emit PoolInitialized(_poolToken, _rewardFunctionId, _timelock);
+    emit PoolInitialized(_poolToken, _rewardFunctionAddress, _timelock);
   }
 
   /**
    * @dev Constructor for direct deployment
    * @param _poolToken Address of the token used for rewards
-   * @param _rewardFunctionId Bytes32 identifier of the reward function (e.g. git commit hash)
+   * @param _rewardFunctionAddress Implementation address of the reward function
    * @param _owner Address that will have DEFAULT_ADMIN_ROLE
    * @param _manager Address that will have MANAGER_ROLE
    * @param _timelock Timestamp when manager withdrawals will be allowed
@@ -177,7 +205,7 @@ contract RewardPool is AccessControl, ReentrancyGuard {
    */
   constructor(
     address _poolToken,
-    bytes32 _rewardFunctionId,
+    address _rewardFunctionAddress,
     address _owner,
     address _manager,
     uint256 _timelock,
@@ -192,13 +220,13 @@ contract RewardPool is AccessControl, ReentrancyGuard {
 
     poolToken = _poolToken;
     isNativeToken = (_poolToken == NATIVE_TOKEN_ADDRESS);
-    rewardFunctionId = _rewardFunctionId;
+    rewardFunctionAddress = _rewardFunctionAddress;
 
     _setTimelock(_timelock);
     _setProtocolFee(_protocolFee);
     _setReserveAddress(_reserveAddress);
 
-    emit PoolInitialized(_poolToken, _rewardFunctionId, _timelock);
+    emit PoolInitialized(_poolToken, _rewardFunctionAddress, _timelock);
   }
 
   /**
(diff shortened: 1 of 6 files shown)


contracts/RewardPool.sol
```diff
@@ -92,6 +96,24 @@ contract RewardPool is AccessControl, ReentrancyGuard {
     uint256 feeAmount,
     uint256 protocolFee
   );
+  event RewardFunctionAddressUpdated(
+    address newRewardFunctionAddress,
+    address previousRewardFunctionAddress
+  );
+  event AddKpi(
```
jeanregisser, line 103: Related to renaming `addKpi`:

```suggestion
  event PeriodKpiUpdated(
```

contracts/RewardPool.sol
```diff
@@ -92,6 +96,24 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+    address indexed user,
+    bytes32 indexed rewardPeriodKey,
+    uint256 amount,
+    uint256 rewardPeriodStart,
+    uint256 rewardPeriodEndExclusive,
```
jeanregisser, line 108: nit: this is sufficient (or even uint48)
```suggestion
    uint64 rewardPeriodStart,
    uint64 rewardPeriodEndExclusive,
```

OpenZeppelin contracts does that too ([example 1](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/53bb34057ed97ea9b36d550f1c2c413ef5b6c6bb/contracts/access/extensions/AccessControlDefaultAdminRules.sol#L56) or [example 2](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/types/Time.sol)).

contracts/RewardPool.sol
```diff
@@ -92,6 +96,24 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+    bytes32 kpiFunctionId
+  );
+  event RewardPeriodProcessed(
+    bytes32 indexed rewardPeriodKey,
+    uint256 rewardPeriodStart,
+    uint256 rewardPeriodEndExclusive,
```
jeanregisser, line 114:
```suggestion
    uint64 rewardPeriodStart,
    uint64 rewardPeriodEndExclusive,
```

contracts/RewardPool.sol
```diff
@@ -120,14 +142,20 @@ contract RewardPool is AccessControl, ReentrancyGuard {
   error AlreadyInitialized();
   error InvalidProtocolFee(uint256 fee);
   error InvalidReserveAddress();
+  error InvalidRewardFunctionAddress();
+  error RewardPeriodInvalid(
+    uint256 rewardPeriodStart,
+    uint256 rewardPeriodEndExclusive
+  );
```
jeanregisser, line 149:
```suggestion
  error InvalidRewardPeriod(
    uint64 rewardPeriodStart,
    uint64 rewardPeriodEndExclusive
  );
```

contracts/RewardPool.sol
```diff
@@ -256,6 +284,141 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+  /**
+   * @dev Allows the owner to process a reward period and distribute rewards
+   * @param rewardPeriodStart Start of the reward period (unix timestamp)
+   * @param rewardPeriodEndExclusive End of the reward period (unix timestamp)
+   * @param totalRewardAmount Total amount of rewards to distribute
+   * @notice Allowed only for address with DEFAULT_ADMIN_ROLE
+   */
+  function processRewardPeriod(
+    uint256 rewardPeriodStart,
+    uint256 rewardPeriodEndExclusive,
```
jeanregisser, line 362:
```suggestion
    uint64 rewardPeriodStart,
    uint64 rewardPeriodEndExclusive,
```

contracts/RewardPool.sol
```diff
@@ -267,49 +430,50 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
-            protocolFee
-          );
-        }
-
-        pendingRewards[reward.user] += reward.amount;
-        totalPendingRewards += reward.amount;
+      if (rewards[i].user == address(0)) revert ZeroAddressNotAllowed(i);
+      if (rewards[i].amount == 0) revert RewardAmountMustBeGreaterThanZero(i);
+      if (rewards[i].idempotencyKey == bytes32(0))
+        revert EmptyIdempotencyKey(i);
```
jeanregisser, line 436: Should we move these checks within `_addReward` to ensure consistency?

contracts/RewardPool.sol
```diff
@@ -393,6 +557,12 @@ contract RewardPool is AccessControl, ReentrancyGuard {
     reserveAddress = _reserveAddress;
   }
 
+  function _setRewardFunctionAddress(address _rewardFunctionAddress) internal {
+    if (_rewardFunctionAddress == address(0))
+      revert InvalidRewardFunctionAddress();
+    rewardFunctionAddress = _rewardFunctionAddress;
+  }
```
jeanregisser, line 564: Do we need this function? I only see it called once from `setRewardFunctionAddress`.

contracts/RewardPool.sol
```diff
@@ -267,49 +430,50 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
 
-        // Old event for backwards compatibility
-        emit AddReward(reward.user, reward.amount, rewardFunctionArgs);
-        emit AddRewardWithIdempotency(
+  function _addReward(
+    RewardData memory reward,
+    uint256[] memory rewardFunctionArgs
+  ) internal {
+    if (!processedIdempotencyKeys[reward.idempotencyKey]) {
```
jeanregisser, line 445: nit: maybe return early so it's a bit more readable?

contracts/RewardPool.sol
```diff
@@ -44,11 +45,14 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
   bool public isNativeToken;
-  bytes32 public rewardFunctionId;
+  address public rewardFunctionAddress;
   uint256 public timelock;
   uint256 public totalPendingRewards;
   mapping(address => uint256) public pendingRewards;
   mapping(bytes32 => bool) public processedIdempotencyKeys;
+  mapping(bytes32 => mapping(address => uint256)) public periodKpis;
+  mapping(bytes32 => address[]) private periodUsers;
+  mapping(bytes32 => bool) public processedPeriods;
```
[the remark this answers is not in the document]
jeanregisser: We could also use [EnumerableSet](https://docs.openzeppelin.com/contracts/5.x/api/utils#EnumerableSet) to have `O(1)` removal instead of currently `O(n)` for `users`.

Though I guess we should still keep in mind that in any case, this will work for when there aren't too many builders (maybe 100, but not 10000?)

contracts/RewardPool.sol
```diff
@@ -92,6 +96,24 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+    uint256 amount,
+    uint256 rewardPeriodStart,
+    uint256 rewardPeriodEndExclusive,
+    bytes32 kpiFunctionId
+  );
+  event RewardPeriodProcessed(
+    bytes32 indexed rewardPeriodKey,
+    uint256 rewardPeriodStart,
+    uint256 rewardPeriodEndExclusive,
+    uint256 totalRewardAmount
```
jeanregisser, line 115: Does the reward function always consume that full amount?
Should we add how many users/builders were processed?

jeanregisser (commented):
Awesome! 🙌

contracts/RewardPool.sol
```diff
@@ -256,6 +289,179 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+    bytes32 kpiFunctionId
+  ) external onlyRole(DEFAULT_ADMIN_ROLE) {
+    bytes32 periodId = _getPeriodId(periodStart, periodEndExclusive);
+
+    PeriodData storage period = _periods[periodId];
+
+    if (period.processedAt != 0) revert PeriodAlreadyProcessed(periodId);
+
+    for (uint256 i = 0; i < kpis.length; i++) {
+      if (kpis[i].referrerAddress == address(0))
```
satish-ravi, line 329: should we rename `referrerAddress` to `user` in `IRewardFunction` to be consistent with addRewards? Although I wonder if that will be confusing because `user` has a different meaning in our scripts.

jeanregisser: Yes I was wondering about this naming too. This contract uses "user" for the referrer. But I guess technically "referrer" is more precised, right?
Would be good to stay consistent and avoid any confusion.

jophish: Hmm, I think my vote would be for `referrer` here in place of `user` -- I think it's easy to get confused and think that `user` is supposed to mean "end user" or "referred user" address instead of "builder/referrer address". Not a huge deal either way, I suppose.

contracts/RewardPool.sol
```diff
@@ -256,6 +289,179 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+        revert ZeroAddressNotAllowed(i);
+
+      period.users.add(kpis[i].referrerAddress);
```
jophish, line 332: The PR description states that `0 kpis are treated as no kpis (can be used for deletion)` -- did we decide that we should keep the `referrerAddress` in the user set even if we update that referrer's KPI value to 0? Does it make sense to remove it from the set, if present, if the given KPI value is 0?

contracts/RewardPool.sol
```diff
@@ -256,6 +289,179 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+
+  /**
+   * @dev Allows the owner to process a reward period and distribute rewards
+   * @param periodStart Start of the reward period (unix timestamp)
+   * @param periodEndExclusive End of the reward period (unix timestamp)
+   * @param totalRewardAmount Total amount of rewards to distribute
+   * @notice Allowed only for address with DEFAULT_ADMIN_ROLE
+   */
+  function processPeriod(
+    uint48 periodStart,
```
jophish, line 354: Probably not a big deal, but there's no way to inspect, just by looking at contract state, what periods have had data uploaded. We can query about particular start/end ranges, but we can't retrieve e.g., a list of periods that are available. With indexing we can look at `KpiUpdated` events, but there's nothing public on the contract that can tell us what's been uploaded.

Not sure if this is a big deal (probably not), but wanted to call it out. It's probably not worth storing the extra state required to return this data.

contracts/RewardPool.sol
```diff
@@ -256,6 +289,179 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+    uint256 numUsersRewarded = 0;
+
+    for (uint32 i = 0; i < rewards.length; i++) {
+      if (rewards[i].reward == 0) continue;
+
+      RewardData memory rewardData = RewardData({
+        user: rewards[i].referrerAddress,
+        amount: rewards[i].reward,
+        idempotencyKey: keccak256(
+          abi.encode(
```
jophish, line 396: I'm not sure it really matters (since any campaign that uses this version of the contract will not be generating idempotency keys off-chain), but the [idempotency key generation in the off-chain scripts](https://github.com/divvi-xyz/divvi-protocol/blob/8ece6ef91f7c9339da73f46cf646b5c8c7569d69/scripts/utils/createSafeTransactionsBatch.ts#L69) differs slightly from the format used here -- I wonder if we should update the off-chain idempotency key generation for consistency's sake. Not a big deal though.

contracts/RewardPool.sol
```diff
@@ -256,6 +289,179 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+            rewards[i].referrerAddress,
+            periodStart,
+            periodEndExclusive
+          )
+        )
+      });
+
+      _addReward(rewardData, rewardFunctionArgs, i);
```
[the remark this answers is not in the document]
jophish: +1 to this -- it actually seems like there are 3 "reward amounts" that could all be different:
* The `totalRewardAmount` (parameter to this method)
* The `totalIssuedRewardAmount` (amount of rewards _actually_ distributed, taking into account idempotency)
* A new `totalPlannedRewardAmount` or something similar; the total amount of rewards planned to be distributed as per the output of `calculateReward` method.

I'm not sure the final one is necessary, but it would be good to update the `_addReward` helper to return a boolean in either case, to ensure `totalssuedRewardAmount` and `numUsersRewarded` are correct.

contracts/RewardPool.sol
```diff
@@ -256,6 +289,179 @@ contract RewardPool is AccessControl, ReentrancyGuard {
...
+  }
+
+  /**
+   * @dev Returns the kpi of a user for a given period
+   * @param periodStart Start of the reward period (unix timestamp)
+   * @param periodEndExclusive End of the reward period (unix timestamp)
+   * @param user Address of the user
+   * @return uint256 kpi
+   */
+  function getPeriodKpi(
```
jophish, line 428: Hmm, right now, if a user (referrer) doesn't have a KPI uploaded for a period, this will return 0; it will also return 0 if a referrer explicitly had a 0 uploaded as a KPI. It will _also_ return 0 if a period doesn't have _any_ data uploaded at all/doesn't exist. 

Should we differentiate between these cases? For example, should we check (using `getPeriodUsers`) if the given user exists in the specified period, and revert with a custom error if they're not present?

contracts/RewardPool.sol
```diff
@@ -41,14 +44,21 @@ contract RewardPool is AccessControl, ReentrancyGuard {
     bytes32 idempotencyKey;
   }
 
+  struct PeriodData {
+    mapping(address => uint256) kpis;
+    EnumerableSet.AddressSet users;
```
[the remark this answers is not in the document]
jophish: Ooh, `EnumerableMap` looks cool. Might be overkill given how we're using these data, and I imagine it's not necessarily more space efficient, but it does seem a bit cleaner than maintaining two data structures ourselves.

contracts/RewardPool.sol
satish-ravi: The description was written before switching to enumerable set / map, where deletes are technically no longer required to keep track of a users list. I didn't like the assumption that `0` kpi means delete. Maybe this is overkill but we could add a `deleteKpi` method to remove

contracts/RewardPool.sol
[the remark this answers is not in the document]
satish-ravi: updated to only include if a reward was issued. I think we should never have `totalIssuedRewardAmount` and `totalPlannedRewardAmount` different. It can only happen if someone manually adds rewards for a period, which I think we should never do if we are using KPIs

contracts/RewardPool.sol
satish-ravi: Probably not a big deal since we ideally should never call `addRewards` if we are using kpis. We can update our off-chain script but one side effect is that active campaigns (tether) has already used the existing format, so maybe fine as is?

jophish: yeah, I think fine as-is!

contracts/RewardPool.sol
jophish: Yeah this is something that we can add in later, seems like it would tie in well with pre-registering the periods before campaign start.

contracts/RewardPool.sol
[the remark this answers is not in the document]
jophish: yeah, given the way idempotency keys are constructed/the inability to process the same period data multiple times, those two values should never differ.

PR #473, opened 2025-09-01
feat(RewardPoolFactory): add reinitializer support for upgrades

jeanregisser:
After deploying the protocol fee changes (see #407 and #472) and having to manually call specific functions on existing factories, I realized this step would be easy to miss in future upgrades. This adds proper [reinitializer](https://docs.openzeppelin.com/contracts/5.x/api/proxy#Initializable-reinitializer-uint64-) support to make the process more foolproof.

**What this adds:**
- `initializeV2()` reinitializer to properly set up new state variables on existing factories
- Enhanced upgrade helpers to automatically call the reinitializer during upgrades
- Updated task with `--call-initialize-v2` flag for explicit control

**How it works:**
- For local upgrades: Just add `--call-initialize-v2` and it handles everything automatically
- For Defender: Outputs the manual call needed after approval (due to their [plugin limitation](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/569))
- `defaultOwner` automatically uses the current factory owner as a sensible default

**Testing:**
Added essential tests to ensure it works correctly and can't be called multiple times.

Now the upgrade process is more explicit and harder to miss the initialization step! ✨


README.md
```diff
@@ -218,6 +218,30 @@ yarn hardhat divvi-registry:upgrade \
     --proxy-address 0x2f5E320698dB89CbefB810Fa19264103d99aAFB1
 ```
 
+To upgrade an existingRewardPoolFactory from V1 to V2 (with protocol fee support), run:
```
jophish, line 221: The full process here also requires us to deploy a new `RewardPool` implementation and call `setImplementation` on the `RewardPoolFactory` to point to the new implementation contract.

jophish (approved):
This is a huge QOL improvement, thanks!

I wonder if it's worth it to have another script here as well, specifically for deploying a new `RewardPool` implementation contract and calling the `setImplementation` method on the factory to point to the new contract as well. Performing this process will often coincide with upgrading the `RewardPoolFactory` contract, but not always, depending on the nature of the `RewardPoolFactory` upgrade.

README.md
jophish: I suppose that we could add the new implementation address as an (optional?) argument to the `initializeV2` method as well, to solve both problems at once?

jophish: I'm planning on upgrading all our RewardPoolFactories today, so I've added this feature in!

README.md
```diff
@@ -218,6 +218,32 @@ yarn hardhat divvi-registry:upgrade \
...
+# With Defender (requires manual reinitializer call after approval)
+yarn hardhat reward-pool-factory:upgrade \
+    --network <NETWORK> \
+    --use-defender \
+    --defender-deploy-salt <SALT> \
+    --proxy-address <PROXY_ADDRESS> \
+    --call-initialize-v2 \
+    --default-protocol-fee <PROTOCOL_FEE> \
+    --default-reserve-address <RESERVE_ADDRESS> \
+    --implementation-address <IMPLEMENTATION_ADDRESS>
```
finnian0826, line 244: This is the same command as above, is that expected? Or should the defender specific stuff be removed from the one above?

finnian0826 (approved):
q about the README command but otherwise looks good!

satish-ravi (approved):

PR #476, opened 2025-09-02
fix(safe-tx-json): Add quotes to values for idempotency version of safe tx json

finnianjs:
Safe UI is expecting the values to be in quotes


jeanregisser (approved):
Thanks! 🚀 

Sorry for not having double checked that part when I made the change 🤦

PR #477, opened 2025-09-03
refactor(contracts): use referrer consistently over user / referrerAddress

satish-ravi:
Follow up based on https://github.com/divvi-xyz/divvi-protocol/pull/462#discussion_r2311590339 

This updates the RewardPool and related contracts to use `referrer` consistently over `user` and `referrerAddress`.

### DIFF
--- a/contracts/RewardPool.sol
+++ b/contracts/RewardPool.sol
@@ -39,7 +39,7 @@ contract RewardPool is AccessControl, ReentrancyGuard {
 
   // Data structures
   struct RewardData {
-    address user;
+    address referrer;
     uint256 amount;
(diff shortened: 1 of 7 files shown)


jeanregisser (approved):
🚀 

Though I'm having second thoughts now 😅 
Maybe we should keep "user" in case we want to support rewards to users too in the future?

satish-ravi:
I was actually torn on the naming as well. Some other potential options are `subject`, `entity`, `beneficiary`. But I think `referrer` is ok for now. This will continue to evolve and if there's a need for user rewards the current kpi -> rewards model may not work. Merging this as is, but happy to change if someone feels strongly

PR #461, opened 2025-08-29
refactor: move some campaign data/config into src/campaigns.ts

silasbw:

PR #451, opened 2025-08-27
feat(redistribution): Add script to redistribute valora rewards

finnian0826:

PR #481, opened 2025-09-04
refactor: remove valora distribution CF

satish-ravi:
To unblock #461, see ENG-607 and #451 for more context

### DIFF
--- a/.knip.json
+++ b/.knip.json
@@ -13,5 +13,6 @@
     "@nomicfoundation/hardhat-ignition",
     "tsc-watch"
   ],
-  "ignoreBinaries": ["gcloud"]
+  "ignoreBinaries": ["gcloud"],
+  "ignore": ["src/campaigns.ts"]
 }


.knip.json
```diff
@@ -13,5 +13,6 @@
     "@nomicfoundation/hardhat-ignition",
     "tsc-watch"
   ],
-  "ignoreBinaries": ["gcloud"]
+  "ignoreBinaries": ["gcloud"],
+  "ignore": ["src/campaigns.ts"]
```
satish-ravi, line 17: temporary, will remove on #461

patches/@safe-global+protocol-kit+6.1.0.patch
satish-ravi, line 1: could leave these in but knip was complaining. We can bring back if we decide to bring back valora redistribution

jophish (approved):
👍

PR #484, opened 2025-09-05
feat: update scripts to support on-chain reward calculation

satish-ravi:
For ENG-596, update periodic kpi job to support both on-chain and off-chain reward calculations.
For on-chain calculations, the script:
- creates a safe tx batch to upload kpis (using `updatePeriodKpis`) and then issue rewards (using `processPeriod`)
- simulates reward calculation by calling the library contract directly to create the rewards.json/csv files

### DIFF
--- a/.github/workflows/calculate-and-upload-KPIs.yaml
+++ b/.github/workflows/calculate-and-upload-KPIs.yaml
@@ -40,6 +40,7 @@ jobs:
             --calculation-timestamp ${{ github.event.inputs.timestamp }} \
             --protocols ${{ github.event.inputs.protocols }} \
             --redis-connection ${{ secrets.REDIS_CONNECTION }}
+            --kpi-function-id ${{ github.sha }}
       - name: Notify Slack on Failure
         if: ${{ failure() || cancelled() }}
         uses: slackapi/slack-github-action@91efab103c0de0a537f72a35f6b8cda0ee76bf0a # v2.1.1
--- a/scripts/createRewardPoolWithKpiTxAndSimulateRewards.ts
+++ b/scripts/createRewardPoolWithKpiTxAndSimulateRewards.ts
@@ -1,0 +1,166 @@
+import yargs from 'yargs'
+import { createUpdateKpiAndProcessRewardsSafeTransactionJSON } from './utils/createSafeTransactionsBatch'
+import { ResultDirectory } from '../src/resultDirectory'
+import {
+  getDivviRewardsExcludedReferrers,
+  ExcludedReferrers,
+} from './utils/divviRewardsExcludedReferrers'
+import fs from 'fs'
+import { parse } from 'csv-parse/sync'
+import { protocols } from './types'
+import { campaigns } from '../src/campaigns'
+import { getReferrerMetricsFromKpi } from './calculateRewards/getReferrerMetricsFromKpi'
+import { getViemPublicClient } from './utils'
+import { rewardFunctionAbi } from '../abis/RewardFunction'
+import { rewardPoolWithKpiAbi } from '../abis/RewardPoolWithKpi'
+
+function parseArgs() {
+  const args = yargs
+    .option('protocol', {
+      description: 'the protocol to calculate rewards for',
+      type: 'string',
+      demandOption: true,
+      choices: protocols,
+    })
+    .option('datadir', {
+      description: 'the directory to store the results',
+      type: 'string',
+      default: 'rewards',
+    })
+    .option('start-timestamp', {
+      alias: 's',
+      description: 'start timestamp',
+      type: 'string',
+      demandOption: true,
+    })
+    .option('end-timestamp', {
+      alias: 'e',
+      description: 'end timestamp',
+      type: 'string',
+      demandOption: true,
+    })
+    .option('reward-amount', {
+      alias: 'r',
+      description: 'the reward amount for this time period in smallest units',
+      type: 'string',
+      demandOption: true,
+    })
+    .option('kpi-function-id', {
+      alias: 'k',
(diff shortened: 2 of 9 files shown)


src/campaigns.ts
```diff
@@ -10,28 +10,48 @@ import { main as calculateRewardsLiskV0 } from '../scripts/calculateRewards/lisk
...
+  startTimestamp: string
+  endTimestampExclusive: string
+}) => Promise<void>
+
+type RewardPeriodWithKpi = {
+  startTimestamp: string
+  endTimestampExclusive: string
+  rewardAmount: string
+  calculateRewards?: never
+  calculateRewardSlices?: CalculateRewardsArgs
```
finnian0826, line 32: Are we still giving out slices? Will this function get used or is it something that can be removed (maybe in a follow-up)?

.github/workflows/calculate-and-upload-KPIs.yaml
```diff
@@ -40,6 +40,7 @@ jobs:
             --calculation-timestamp ${{ github.event.inputs.timestamp }} \
             --protocols ${{ github.event.inputs.protocols }} \
             --redis-connection ${{ secrets.REDIS_CONNECTION }}
+            --kpi-function-id ${{ github.sha }}
```
finnian0826, line 43: Just to make sure I understand, this will be the commit hash of the lates commit on main right? As compared to what we include for reward pools which is the commit hash that added the reward function.

scripts/utils/createSafeTransactionsBatch.ts
```diff
@@ -124,3 +125,74 @@ export const createAddRewardSafeTransactionJSON = ({
...
+        contractInputsValues: processPeriodMethodInputs,
+      },
+    ],
+  }
+
+  // Create directory if it doesn't exist
+  mkdirSync(dirname(filePath), { recursive: true })
+  writeFileSync(filePath, JSON.stringify(transactionsBatch, null, 2) + '\n', {
+    encoding: 'utf-8',
+  })
```
finnian0826, line 197: should we create something like `resultDirectory.writeRewards(...)` but for safe tx json? That way we only have to pass around the resultDirectory and then matches all the other file writes we do.

scripts/createRewardPoolWithKpiTxAndSimulateRewards.ts
```diff
@@ -0,0 +1,166 @@
...
+    address: rewardFunctionAddress,
+    abi: rewardFunctionAbi,
+    functionName: 'calculateReward',
+    args: [kpis, BigInt(rewardAmount)],
+  })
+
+  const rewardsWithMetadata = rewards.map((reward) => ({
+    ...reward,
+    totalKpi: referrerKpis[reward.referrer],
+    referralCount: referrerReferrals[reward.referrer],
```
finnian0826, line 154: this excludes some metadata that could be present in a campaigns kpi file, like kpi breakdown by chain. For scout-game we had some custom code in reward calculation to include that aggregated metadata per referrer, but for tether we didn't add it so only have that metadata in kpi. Is there some way we can aggregate all the metadata per referrer that exists in a campaigns kpi calculation? Or are these two values all we really care about for now?

scripts/createRewardPoolWithKpiTxAndSimulateRewards.ts
```diff
@@ -0,0 +1,166 @@
...
+    protocol: args.protocol,
+    kpiFunctionId: args['kpi-function-id'],
+  }
+}
+
+export async function main(args: ReturnType<typeof parseArgs>) {
+  const startTimestamp = new Date(args.startTimestamp)
+  const endTimestampExclusive = new Date(args.endTimestampExclusive)
+  const resultDirectory = args.resultDirectory
+  const rewardAmount = args.rewardAmount
```
finnian0826, line 100:
```suggestion
  const rewardAmount = BigInt(args.rewardAmount)
```
to avoid having to cast it twice below

scripts/uploadCurrentPeriodKpis.ts
```diff
@@ -165,6 +172,26 @@ export async function uploadCurrentPeriodKpis(
...
+        kpiFunctionId: args.kpiFunctionId,
+      })
+      const rewardsFilePathCsv = join(outputDir, 'rewards.csv')
+      const rewardsFilePathJson = join(outputDir, 'rewards.json')
+      const safeTransactionsJson = join(outputDir, 'safe-transactions.json')
+      campaignFilePaths.push(
+        rewardsFilePathCsv,
+        rewardsFilePathJson,
+        safeTransactionsJson,
+      )
```
finnian0826, line 192: this is duplicated below, could do the file path creation outside of the conditionals to make sure names match in case someone updates one (unless we would want them to be different), and then only have the push inside the conditionals

finnian0826 (approved):
looks good to me!

.github/workflows/calculate-and-upload-KPIs.yaml
satish-ravi: yes thats right.

scripts/createRewardPoolWithKpiTxAndSimulateRewards.ts
satish-ravi: I think those would be hard to do in a generic manner as the metadata is typed as `Record<string, unknown>`, we can't really aggregate them without knowing the type of the values. We could do a custom aggregator per campaign but probably not needed now.

scripts/uploadCurrentPeriodKpis.ts
satish-ravi: updated to use paths from resultDirectory

PR #513, opened 2025-09-24
refactor(celo-pg): speciy KPI files to use for stage calculation

silasbw:
Can be file path or HTTPS url

### DIFF
--- a/src/campaigns.ts
+++ b/src/campaigns.ts
@@ -403,7 +403,7 @@ export const campaigns: Campaign[] = [
             endTimestampExclusive,
             rewardAmount: '25000',
             excludedReferrers: excludedReferrersFromCeloPGS1[0],
-            previousResultDirectories: [],
+            previousKpiFiles: [],
           })
         },
       },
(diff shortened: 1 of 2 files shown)


src/campaigns.ts
```diff
@@ -421,13 +421,8 @@ export const campaigns: Campaign[] = [
...
             excludedReferrers: {},
-            previousResultDirectories: [
-              new ResultDirectory({
-                datadir: 'rewards',
-                name: 'celo-pg-s1',
-                startTimestamp: new Date('2025-08-26T00:00:00Z'),
-                endTimestampExclusive: new Date('2025-09-09T00:00:00Z'),
-              }),
+            previousKpiFiles: [
+              'https://storage.googleapis.com/divvi-campaign-data-production/kpi/celo-pg-s1/2025-08-26T00:00:00.000Z_2025-09-09T00:00:00.000Z/kpi.json',
```
silasbw, line 425: FYI, i'll make this a little nicer and fill our the other previousKpiFile properties in a follow-on PR

src/campaigns.ts
```diff
@@ -446,7 +441,7 @@ export const campaigns: Campaign[] = [
             endTimestampExclusive,
             rewardAmount: '25000',
             excludedReferrers: {},
-            previousResultDirectories: [],
+            previousKpiFiles: [''],
```
kathaypacific, line 444: should this be just an empty array? i _think_ `readKpiFile` expects non empty strings, but maybe you have plans for this in a follow up

kathaypacific (approved):

